PR #1, opened 2019-05-12
First version of a lang tester

ltratt:
This is a heavy rewrite of the preview from a day or two back, though the external API is mostly unchanged. The good news is that this is powerful enough to deal with both Rust and SOM tests, and to do so requiring very little code from the user of the library.

What has changed significantly since you last saw things is the internal code and the format of the tests themselves. I moved from a custom parser to YAML, but that was horrible: multi-line, nested YAML simply doesn't work as I expect, and the Rust YAML parser gives confusing error messages. I've thus moved to TOML. Is this a good idea? I'm wondering if the best syntax would, regrettably, be a custom indentation (almost YAML, but not quite) one:

```
// Compiler:
//     status: success
//     stderr:
//         warning: unused variable: `x`
//         ...unused_var.rs:13:9
//         ...
//
// Run-time:
//     status: success
//     stdout: Hello world
```

Feedback on that aspect welcome!

If you want to try this out, the easiest thing to do is run:

```sh
$ cargo run --example=rust_lang_tester
   Compiling lang_tester v0.1.0 (/home/ltratt/scratch/softdev/lang_tester)
    Finished dev [unoptimized + debuginfo] target(s) in 2.48s
     Running `target/debug/examples/rust_lang_tester`

running 3 tests
test lang_tests::no_main ... ok
test lang_tests::unknown_var ... ok
test lang_tests::unused_var ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

and then play with the test files in `examples/rust_lang_tester` ([here's the relevant file in the diff](https://github.com/softdevteam/lang_tester/pull/1/files#diff-31c7fec71a87b4badb3b56d964f5d675)), breaking them and so on.

Until we've confirmed what the test format should be, I'm going to hold off documenting the top-level of the library. But I don't think we're *too* far away from having something useful here.

### DIFF
--- a/examples/rust_lang_tester/lang_tests/no_main.rs
+++ b/examples/rust_lang_tester/lang_tests/no_main.rs
@@ -1,0 +1,10 @@
+// Compiler:
+//   status: error
+//   stderr:
+//     error[E0601]: `main` function not found in crate `no_main`
+//       |
+//       = note: consider adding a `main` function to `examples/rust_lang_tester/lang_tests/no_main.rs`
+//
+//     error: aborting due to previous error
+//
+//     For more information about this error, try `rustc --explain E0601`.
--- a/examples/rust_lang_tester/lang_tests/unknown_var.rs
+++ b/examples/rust_lang_tester/lang_tests/unknown_var.rs
@@ -1,0 +1,10 @@
+// Compiler:
+//   status: error
+//   stderr:
+//     error[E0425]: cannot find value `x` in this scope
+//      ...unknown_var.rs:9:20
+//      ...
+
+fn main() {
+    println!("{}", x);
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,106 @@
+// Copyright (c) 2019 King's College London created by the Software Development Team
+// <http://soft-dev.org/>
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0>, or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, or the UPL-1.0 license <http://opensource.org/licenses/UPL>
+// at your option. This file may not be copied, modified, or distributed except according to those
+// terms.
+
+//! This crate provide a simple language testing framework designed to help when you are testing
+//! things like compilers and virtual machines. It allows users to embed simple tests for process
+//! success/failure and for stderr/stdout inside a source file. It is loosely based on the
+//! [`compiletest_rs`](https://crates.io/crates/compiletest_rs) crate, but is much simpler (and
+//! hence sometimes less powerful), and designed to be used for testing non-Rust languages too.
+//!
+//! For example, a Rust language tester, loosely in the spirit of
+//! [`compiletest_rs`](https://crates.io/crates/compiletest_rs), looks as follows:
+//!
+//! ```rust
+//! use std::{path::PathBuf, process::Command};
+//!
+//! use lang_tester::LangTester;
+//! use tempdir::TempDir;
+//!
+//! fn main() {
+//!     // We use rustc to compile files into a binary: we store those binary files into `tempdir`.
+//!     // This may not be necessary for other languages.
+//!     let tempdir = TempDir::new("rust_lang_tester").unwrap();
+//!     LangTester::new()
+//!         .test_dir("examples/rust_lang_tester/lang_tests")
+//!         // Only use files named `*.rs` as tests.
+//!         .test_file_filter(|p| p.extension().unwrap().to_str().unwrap() == "rs")
+//!         // Extract the first sequence of commented line(s) as the test.
+//!         .test_extract(|s| {
+//!             Some(
+//!                 s.lines()
+//!                     // Skip non-commented lines at the start of the file.
+//!                     .skip_while(|l| !l.starts_with("//"))
+//!                     // Extract consecutive commented lines.
+//!                     .take_while(|l| l.starts_with("//"))
+//!                     .map(|l| &l[2..])
+//!                     .collect::<Vec<_>>()
+//!                     .join("\n"),
+//!             )
+//!         })
+//!         // We have two test commands:
+//!         //   * `Compiler`: runs rustc.
+//!         //   * `Run-time`: if rustc does not error, and the `Compiler` tests succeed, then the
+//!         //     output binary is run.
+//!         .test_cmds(move |p| {
+//!             // Test command 1: Compile `x.rs` into `tempdir/x`.
+//!             let mut exe = PathBuf::new();
+//!             exe.push(&tempdir);
+//!             exe.push(p.file_stem().unwrap());
+//!             let mut compiler = Command::new("rustc");
+//!             compiler.args(&["-o", exe.to_str().unwrap(), p.to_str().unwrap()]);
+//!             // Test command 2: run `tempdir/x`.
+//!             let runtime = Command::new(exe);
+//!             vec![("Compiler", compiler), ("Run-time", runtime)]
+//!         })
+//!         .run();
+//! }
+//! ```
+//!
+//! This defines a lang tester that uses all `*.rs` files in a given directory as tests, running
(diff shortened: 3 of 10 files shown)


ltratt:
I should have opened this is a draft pull request. Unfortunately, you can't go from "ready" to "draft" (yet https://github.community/t5/How-to-use-Git-and-GitHub/Feature-Request-Switch-from-ready-to-draft-in-pull-requests/td-p/19107). Anyway, don't merge it yet please ;)

examples/rust_lang_tester/lang_tests/no_main.rs
```diff
@@ -0,0 +1,11 @@
+// [Compiler]
+// status = "error"
+// stderr = """
```
vext01: The triple-quoting required is a minor annoyance, but it's not really enough to sufficiently upset me though.

examples/rust_lang_tester/lang_tests/unknown_var.rs
```diff
@@ -0,0 +1,11 @@
+// [Compiler]
+// status = "error"
```
jacob-hughes: Are the only possibilities for `status` `"error"` and `"success"`? In which case, would it make more sense for this to be `success = true | false`?

src/lib.rs
```diff
@@ -0,0 +1,464 @@
...
+                                });
+                            }
+                            _ => panic!("TOML entry {} must contain a table.", cmd_name),
+                        },
+                    }
+                }
+            }
+            _ => panic!("Test TOML must be a table."),
+        }
+
```
jacob-hughes: Whitespace?

examples/rust_lang_tester/lang_tests/unknown_var.rs
ltratt: This is one of those things I'm a bit unsure about. We *could* make this test for Unix error returns (which would be integers), but I reckon 99% of the time you just care about "succeeded" or "errored". But, yes, I tend to agree that "true", "false" is better.

src/lib.rs
ltratt: That one was intentional, just to differentiate it from the massive `match` above. But I don't feel strongly about it. Since it looks like I'm going to rewrite the parser, I'm not sure it's too important.

examples/rust_lang_tester/lang_tests/unknown_var.rs
vext01: We've just had an in-person conversation:

If we were going to be super-anal, we should be checking the exit status against an integer. Laurie isn't keen on doing that now, but if we leave the keyword as "status", then that allow us to at-least implement later.

Put differently, if we commit to `success=true/false`, we can't later add `success=6`, which makes little sense.

So I'd say, leave it as it is.

jacob-hughes:
I've downloaded and run the test suite against the Rust language test runner example. It looks like it will find the first comment and ignore duplicate uses of `status`; `stdout`; etc found in comments later on in the file. This does seem sensible to me. Is this the intended behaviour?

One other thing I noticed that we have to be a little careful about: comments can't always be at the top of the file in Rust. For example, in the GC we might need to turn on the forked compiler's GC feature flag (`#![feature(gc)]`). These need to be at the top of the file, and when I added something similar to `no_main.rs`, comment parsing broke  (`ignored: (test string is empty)`)

examples/rust_lang_tester/lang_tests/unknown_var.rs
jacob-hughes: Yes that sounds reasonable to me

examples/rust_lang_tester/lang_tests/no_main.rs
jacob-hughes: I suppose an alternative approach, which is used in Rust's `compiletest` framework, is to have separate files for the resulting program's `stdout` and `stderr`. For example, `no_main.rs` would have a corresponding `no_main.rs.stdout` etc.. and these files are then read and compared to the actual output.

The benefit of this is we need less (or maybe even none at all if we're clever) language specific pattern matching code from the user. The problem is I can't think of a clean way to separate these on a per-command basis. If we're testing both compilation _and_ runtime, we'd end up with around 5 files per single test which might be overkill. Rust solves this by splitting tests into separate `compile_*` and `run_*` directories, but at this point I wonder if we'd just end up reinventing the wheel.

vext01: I already mentioned this approach to Laurie, and he's not up for it ;)

ltratt: Yeah, I definitely don't want to do this, because the API would become much more complex.

ltratt:
>I've downloaded and run the test suite against the Rust language test runner example. It looks like it will find the first comment and ignore duplicate uses of status; stdout; etc found in comments later on in the file. This does seem sensible to me. Is this the intended behaviour?

It's whatever `rust_lang_tester` decides it wants to do (in the closure passed to `test_extract`). At the moment it does what you say above. The `lang_tester` library itself has absolutely no opinions on this.

>One other thing I noticed that we have to be a little careful about: comments can't always be at the top of the file in Rust. For example, in the GC we might need to turn on the forked compiler's GC feature flag (#![feature(gc)]). These need to be at the top of the file, and when I added something similar to no_main.rs, comment parsing broke (ignored: (test string is empty))

Does the `#![feature...]` really have to be before the comment?

ltratt:
@jacob-hughes I've now pushed a first attempt at a parser for the new YAML-ish (but not YAML!) syntax. It's only been lightly tested, so feel free to report problems!

src/lib.rs
```diff
@@ -0,0 +1,538 @@
...
+                    e.insert(test);
+                }
+            }
+        }
+        tests
+    }
+
+    /// Check for the case where the user has a test such as:
+    ///   [Test]
+    /// but test_cmds doesn't have any test called "Test". This is almost certainly a bug, in the
```
jacob-hughes: I'm not sure I understand this part. What does it mean to have a "test such as `[Test]`"

jacob-hughes: Sorry ignore me. I get it now... I just misread the comment :man_facepalming:

jacob-hughes:
>Does the #![feature...] really have to be before the comment? 

Yes, unfortunately.

I actually don't mind the new syntax. I'm not a fan of YAML with all its 100 variants for representing basically the same thing. This approach seems to fit our use-case quite nicely. 

Regarding the new test parser, it LGTM I can't any glaring mistakes.

src/lib.rs
ltratt: Sorry, it's also using the old syntax...

ltratt:
>> Does the #![feature...] really have to be before the comment? 
>Yes, unfortunately.

I'm wondering if I should rename `examples/rust_lang_tester/` to something like `examples/simple_rust_lang_tester` to make clear it's not a drop-in replacement for rustc's system. Any thoughts?

ltratt:
https://github.com/softdevteam/lang_tester/pull/1/commits/df1879b203db8a440bd71bb808fe0fb78e60dc25 is a big, but boring, code shuffle.

https://github.com/softdevteam/lang_tester/pull/1/commits/f97db76483522c088a26459b428dd32e4d64fef6 is a tiny change that should make @jacob-hughes happy for Rust testing.

I think we're at the point where I can finish the docs off and then we can merge?

ltratt:
With that last commit I think I am more-or-less done, at least for this PR. Comments welcome.

jacob-hughes:
I'm happy with the documentation too. Do you want to squash?

ltratt:
I do! Squashed.

jacob-hughes:
> I should have opened this is a draft pull request. Unfortunately, you can't go from "ready" to "draft" (yet https://github.community/t5/How-to-use-Git-and-GitHub/Feature-Request-Switch-from-ready-to-draft-in-pull-requests/td-p/19107). Anyway, don't merge it yet please ;)

I'm guessing this no longer applies? If so, I'll merge

ltratt:
Good question! Yes, I think this can be merged now.

PR #2, opened 2019-05-14
Better fuzzy matching

ltratt:
After breaking the fuzzy matcher out into its own file (it's stand alone, so doing so helps make `tester.rs` feel a little less intimidating), this fixes a bug in the current fuzzy matcher (https://github.com/softdevteam/lang_tester/commit/c9317f508e6ad0c50bfb553060f663fdd960312f) and then allows `...` to be used at the start and/or end of the line (https://github.com/softdevteam/lang_tester/commit/65d39d86b249b5fbb43dfb1d8f51549275c64076). This doesn't hugely affect code complexity, but does make it easier to think about from an end-user's perspective IMHO>

### DIFF
--- a/src/fuzzy.rs
+++ b/src/fuzzy.rs
@@ -26,14 +26,8 @@ pub(crate) fn match_vec(plines: &Vec<&str>, s: &str) -> bool {
             if plines[pi] == WILDCARD {
                 panic!("Can't have '{}' on two consecutive lines.", WILDCARD);
             }
-            while si < slines.len() {
-                if match_line(plines[pi], slines[si]) {
-                    break;
-                }
+            while si < slines.len() && !match_line(plines[pi], slines[si]) {
                 si += 1;
-            }
-            if si == slines.len() {
-                return false;
             }
         } else if match_line(plines[pi], slines[si]) {
             pi += 1;
@@ -42,13 +36,25 @@ pub(crate) fn match_vec(plines: &Vec<&str>, s: &str) -> bool {
             return false;
         }
     }
-    true
+    (pi == plines.len() && si == slines.len())
+        || (pi + 1 == plines.len() && plines[pi] == WILDCARD && si == slines.len())
 }
 
 /// Does the line `s` match the pattern `p`? Note that both strings are expected to be trimed
 /// before being passed to this function.
 fn match_line(p: &str, s: &str) -> bool {
-    (p.starts_with(WILDCARD) && s.ends_with(&p[WILDCARD.len()..])) || p == s
+    let sww = p.starts_with(WILDCARD);
+    let eww = p.ends_with(WILDCARD);
+    if sww && eww {
+        s.find(&p[WILDCARD.len()..p.len() - WILDCARD.len()])
+            .is_some()
+    } else if sww {
+        s.ends_with(&p[WILDCARD.len()..])
+    } else if eww {
+        s.starts_with(&p[..p.len() - WILDCARD.len()])
+    } else {
+        p == s
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


ltratt:
https://github.com/softdevteam/lang_tester/pull/2/commits/3158f35f5f6c2ca82e9f86daaf57fee08a5cf894 fixes a minor bug I reintroduced in an earlier commit in this PR, and does some tidying up at the same time. It would have been part of the original PR if I'd been thinking properly about it...

src/fuzzy.rs
```diff
@@ -48,6 +45,10 @@ pub(crate) fn match_vec(plines: &Vec<&str>, s: &str) -> bool {
     true
 }
 
+fn match_line(p: &str, s: &str) -> bool {
+    (p.starts_with(WILDCARD) && s.ends_with(&p[WILDCARD.len()..])) || p == s
```
jacob-hughes: And we ensure that `p` and `s` are trimmed before they enter `match_vec`, so we don't need to handle whitespace either side of the line?

src/fuzzy.rs
```diff
@@ -26,27 +26,33 @@ pub(crate) fn match_vec(plines: &Vec<&str>, s: &str) -> bool {
...
+    (pi == plines.len() && si == slines.len())
+        || (pi + 1 == plines.len() && plines[pi] == WILDCARD && si == slines.len())
 }
 
 fn match_line(p: &str, s: &str) -> bool {
-    (p.starts_with(WILDCARD) && s.ends_with(&p[WILDCARD.len()..])) || p == s
+    let sww = p.starts_with(WILDCARD);
+    let eww = p.ends_with(WILDCARD);
+    if sww && eww {
+        s.find(&p[WILDCARD.len()..p.len() - WILDCARD.len()])
```
jacob-hughes: Controversially, I would prefer this written as `WILDCARD.len()..(p.len() - WILDCARD.len())`. My brain can chunk the expression into the correct precedence rules faster.

src/lib.rs
```diff
@@ -97,8 +97,11 @@
 //! Lines not mentioned are not tested: for example, the above file does not state whether the
 //! `Compiler`s `stdout` should have content or not (but note that the line `stdout:` on its own
 //! would state that the `Compiler` should have no content at all). `stderr`/`stdout` tests can use
-//! `...` as a simple wildcard: if a line consists solely of `...`, it means either "match zero or
-//! more lines"; if a line begins with `...`, it means "match the remainder of the line only".
+//! `...` as a simple wildcard: if a line consists solely of `...`, it means "match zero or
+//! more lines"; if a line begins with `...`, it means "match the remainder of the line only";
+//! if a line ends with `...`, it means "match the start of the line only". A line may start and
+//! end with `...`. `stderr`/`stdout` matches ignore leading/trailing whitespace and newlines, but
+//! are case sensitive.
```
jacob-hughes: Lets add an example. It's not too hard to understand but it's the first thing people jump to when they see an intimidating block of documentation :P

src/fuzzy.rs
ltratt: Correct. I should probably make that clear though.

src/fuzzy.rs
ltratt: I think rustfmt will delete the extra brackets?

jacob-hughes: Oh yes I didn't think of that. Maybe it will. In which case, nevermind. 

Although if that's true I don't really agree that the best formatted expression is the one with minimal parenthesis. It forces people to remember the order of precedence (which can be really annoying for bitwise operators that aren't used frequently)

src/fuzzy.rs
jacob-hughes: Yeah a doc comment on the function wouldn't hurt

jacob-hughes:
Other than those minor comments, this PR LGTM

src/fuzzy.rs
ltratt: I agree generally (I can never remember precedence order in general) but I think it's clear that `..` must have an exceptionally low precedence!

ltratt:
Hopefully that new commit addresses your concerns.

jacob-hughes:
Yes it does. Please squash!

ltratt:
Squashed.

PR #5, opened 2019-05-17
Allow integer codes to be used in the "status" field.

### LINKED ISSUE #4: How I can check return value from main function?
Let's say we have this code in some file:
```go
pub func main() i32 {
     var variable = awesome_function();
     return variable;
}
```
How I can check this return using lang tester?

ltratt:
Fixes #4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,10 @@ fn main() {
 Tests use a two-level indentation syntax: the outer most level of indentation
 defines a command name (multiple command names can be specified, as in the
 above); each command name can then define tests for one or more of `status:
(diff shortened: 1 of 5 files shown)


jacob-hughes:
LGTM - I did wonder why you didn't use a `u8`, especially considering that most shells I've ever seen only deal with error codes from 0-255. I'm mistaken though, according to POSIX it should be a signed 32 bit integer!

This should work with windows too, which AFAIK uses a 32 bit _unsigned_ integer instead.

ltratt:
I just copied the type from Rust's `process::exit` without thinking ;) If you're happy, this can be merged?

PR #6, opened 2019-05-17
Various improvements suggested by Clippy.

ltratt:
And turn off some of the more annoying ones that would make the code more complex for no gain...

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -77,7 +77,7 @@ pub(crate) fn parse_tests<'a>(test_str: &'a str) -> HashMap<String, Test<'a>> {
     tests
 }
 
-fn indent_level(lines: &Vec<&str>, line_off: usize) -> usize {
+fn indent_level(lines: &[&str], line_off: usize) -> usize {
     lines[line_off]
         .chars()
         .take_while(|c| c.is_whitespace())
@@ -85,7 +85,7 @@ fn indent_level(lines: &Vec<&str>, line_off: usize) -> usize {
 }
 
 /// Turn a line such as `key: val` into its separate components.
-fn key_val<'a>(lines: &Vec<&'a str>, line_off: usize, indent: usize) -> (&'a str, &'a str) {
+fn key_val<'a>(lines: &[&'a str], line_off: usize, indent: usize) -> (&'a str, &'a str) {
     let line = lines[line_off];
     let key_len = line[indent..]
         .chars()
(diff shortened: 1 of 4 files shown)


src/tester.rs
```diff
@@ -207,7 +208,7 @@ impl<'a> LangTester<'a> {
             for (cmd_name, mut cmd) in cmd_pairs {
                 let output = cmd
                     .output()
-                    .expect(&format!("Couldn't run command {:?}.", cmd));
+                    .unwrap_or_else(|_| panic!(format!("Couldn't run command {:?}.", cmd)));
```
jacob-hughes: Why is this better?

vext01: I also prefer how it was before :P

src/parser.rs
```diff
@@ -77,15 +77,15 @@ pub(crate) fn parse_tests<'a>(test_str: &'a str) -> HashMap<String, Test<'a>> {
     tests
 }
 
-fn indent_level(lines: &Vec<&str>, line_off: usize) -> usize {
+fn indent_level(lines: &[&str], line_off: usize) -> usize {
```
jacob-hughes: I've seen in the past Rust people stating that passing a slice over a `Vec` here more idiomatic. I can't seem to find the find the a link to the reasoning anymore though. Do you know why this is better?

src/tester.rs
ltratt: It means that the `format` stuff isn't executed *unless* there's been a failure.

src/parser.rs
ltratt: It's more flexible: you can do things like `&vec[n..m]` and so on. Is it useful here? Not really. But it's not unuseful, and it's slightly fewer characters ;)

src/tester.rs
jacob-hughes: Ah. That's beautifully non-obvious. I'd expect expect to do that.

ltratt: `expect` is a normal function, so arguments are eagerly evaluated. We can't blame `expect` here :)

vext01: Interesting! In that case, I retract my comment.

Shame it makes the code less readable.

src/parser.rs
vext01: I guess a slice is more generic? It could be from a Vec or and array?

jacob-hughes: Ok I'm convinced.

src/tester.rs
jacob-hughes: I see that now yeah, that's why it takes a closure.

vext01:
Yes, sorry, go ahead!

PR #7, opened 2019-05-20
Allow the user to specify which tests they want to run on the command line

ltratt:
This means that doing things like:

```
cargo test a
```

only runs tests with "a" in their name.

Since this feature has to read command-line parameters, which might not be appropriate in all places, it can be switched off by specifying `use_cmdline_args(false)` (the default is `true`).

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -9,12 +9,14 @@
 
 use std::{
     collections::{hash_map::HashMap, HashSet},
+    env,
     fs::read_to_string,
     io::{self, Write},
     path::{Path, PathBuf},
     process::{self, Command},
 };
 
+use getopts::Options;
 use termcolor::{Color, ColorChoice, ColorSpec, StandardStream, WriteColor};
 use walkdir::WalkDir;
 
@@ -25,6 +27,8 @@ pub struct LangTester<'a> {
     test_file_filter: Option<Box<Fn(&Path) -> bool>>,
     test_extract: Option<Box<Fn(&str) -> Option<String>>>,
     test_cmds: Option<Box<Fn(&Path) -> Vec<(&str, Command)>>>,
+    use_cmdline_args: bool,
+    cmdline_filters: Option<Vec<String>>,
 }
 
 impl<'a> LangTester<'a> {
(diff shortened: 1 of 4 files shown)


src/tester.rs
```diff
@@ -144,6 +150,28 @@ impl<'a> LangTester<'a> {
         self
     }
 
+    /// If set to `true`, this reads arguments from `std::env::args()` and interprets them in the
+    /// same way as normal cargo test files. For example if you have test "ab" and "cd" but only want
```
jacob-hughes: typo, should this be "If you have **tests** "ab" and..." ?

src/tester.rs
```diff
@@ -144,6 +150,28 @@ impl<'a> LangTester<'a> {
...
+    /// As this suggests, a simple substring search is used to decide which tests to run.
+    ///
+    /// You can get help on `lang_tester`'s options:
+    ///
+    /// ```sh
+    /// $ <test bin> --help
+    /// ```
+    ///
+    /// This option defaults to `true`.
+    pub fn use_cmdline_args(&'a mut self, use_cmdline_args: bool) -> &'a mut Self {
```
jacob-hughes: Why might I want to want to switch this off? If I don't care to filter the tests, I won't run `test` with a filter. I'm guessing maybe in case there is ambiguity with other `test` command line arguments?

src/tester.rs
ltratt: Oops!

src/tester.rs
ltratt: Someone might run this as part of some non-cargo setup, and not want `lang_tester` to look at the command-line arguments. I don't have a concrete use-case for this.

jacob-hughes: Fair enough

src/tester.rs
```diff
@@ -157,24 +185,60 @@ impl<'a> LangTester<'a> {
...
     pub fn run(&mut self) {
         self.validate();
-        let test_files = self.test_files();
+        if self.use_cmdline_args {
+            let args: Vec<String> = env::args().collect();
+            let matches = Options::new()
+                .optflag("h", "help", "")
+                .parse(&args[1..])
+                .unwrap_or_else(|_| usage());
+            if matches.opt_present("h") {
```
jacob-hughes: Does this match both the long and short name?

ltratt: Yes.

jacob-hughes:
LGTM. Please squash.

ltratt:
Squashed.

PR #8, opened 2019-05-20
Parallel tests

ltratt:
This makes `lang_tester` run tests in parallel (one job per CPU core). This seriously speeds things up -- yksom's tiny test suite of 6 language tests by over 2x on my 2 core laptop with hyper-threading on (and a mere ~1.8x with hyper-threading off).

https://github.com/softdevteam/lang_tester/commit/9cc27f7abbeab154d09af9b911bb701944cbf828 does the parallelisation in a simple but effective way. https://github.com/softdevteam/lang_tester/commit/da955eef58bb34de2399f764f73434551a1dded6 and https://github.com/softdevteam/lang_tester/commit/2e1dbe66dd3661f3b5fbdae72237cceb44031d8e are semi-mechanical commits which prepare the ground for doing "full and proper" parallisation in https://github.com/softdevteam/lang_tester/commit/6159172091ea9c0d97d9cac97ec3b65da73ed72e.

### DIFF
--- a/src/fuzzy.rs
+++ b/src/fuzzy.rs
@@ -11,7 +11,7 @@ const WILDCARD: &str = "...";
 
 /// Does `s` conform to the fuzzy pattern `pattern`? Note that `plines` is expected not to start or
 /// end with blank lines, and each line is expected to be `trim`ed.
-pub(crate) fn match_vec(plines: &[&str], s: &str) -> bool {
+pub(crate) fn match_vec(plines: &Vec<String>, s: &str) -> bool {
     let slines = s.trim().lines().map(|x| x.trim()).collect::<Vec<_>>();
 
     let mut pi = 0;
@@ -26,10 +26,10 @@ pub(crate) fn match_vec(plines: &[&str], s: &str) -> bool {
             if plines[pi] == WILDCARD {
                 panic!("Can't have '{}' on two consecutive lines.", WILDCARD);
             }
-            while si < slines.len() && !match_line(plines[pi], slines[si]) {
+            while si < slines.len() && !match_line(&plines[pi], slines[si]) {
                 si += 1;
             }
-        } else if match_line(plines[pi], slines[si]) {
+        } else if match_line(&plines[pi], slines[si]) {
             pi += 1;
             si += 1;
         } else {
@@ -64,7 +64,7 @@ mod tests {
     #[test]
     fn test_match_vec() {
         fn match_vec_helper(p: &str, s: &str) -> bool {
-            match_vec(&p.lines().collect::<Vec<_>>(), s)
+            match_vec(&p.lines().map(|x| x.to_owned()).collect::<Vec<_>>(), s)
         }
         assert!(match_vec_helper("", ""));
         assert!(match_vec_helper("a", "a"));
--- a/src/fuzzy.rs
+++ b/src/fuzzy.rs
@@ -11,7 +11,7 @@ const WILDCARD: &str = "...";
 
 /// Does `s` conform to the fuzzy pattern `pattern`? Note that `plines` is expected not to start or
 /// end with blank lines, and each line is expected to be `trim`ed.
-pub(crate) fn match_vec(plines: &Vec<String>, s: &str) -> bool {
+pub(crate) fn match_vec(plines: &[&str], s: &str) -> bool {
     let slines = s.trim().lines().map(|x| x.trim()).collect::<Vec<_>>();
 
     let mut pi = 0;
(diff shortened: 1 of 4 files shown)


ltratt:
Oops, https://github.com/softdevteam/lang_tester/pull/8/commits/adb30dbc473a35b17d07a7330e36fc861f0c9f6b is something I overlooked.

src/tester.rs
```diff
@@ -439,15 +343,162 @@ fn write_with_colour(s: &str, colour: Color) {
     stderr.reset().ok();
 }
 
-fn output_failed() {
```
jacob-hughes: Is the reason for removing these function calls and writing straight to `stderr` related to problems with parallelism?

src/tester.rs
```diff
@@ -439,15 +343,162 @@ fn write_with_colour(s: &str, colour: Color) {
...
-    process::exit(1);
+/// Run every test in `test_files`, returning a tuple `(failures, num_ignored)`.
+fn run_tests(
+    test_files: Vec<PathBuf>,
+    inner: Arc<LangTesterInner>,
+) -> (Vec<(String, TestFailure)>, usize) {
+    let failures = Arc::new(Mutex::new(Vec::new()));
+    let mut num_ignored = 0;
+    let pool = ThreadPool::new(num_cpus::get());
+    'b: for p in test_files {
```
jacob-hughes: Shall we remove the loop label if it's not used? (I'm also really not a fan of how it looks like a lifetime!)

src/tester.rs
jacob-hughes: Ah, I've seen further down that you acquire a lock on `stderr`, so I'm guessing it's to ensure this happens at the correct place?

src/tester.rs
ltratt: Good point!

ltratt:
The new commit should address your concerns with luck.

src/fuzzy.rs
```diff
@@ -11,7 +11,7 @@ const WILDCARD: &str = "...";
 
 /// Does `s` conform to the fuzzy pattern `pattern`? Note that `plines` is expected not to start or
 /// end with blank lines, and each line is expected to be `trim`ed.
-pub(crate) fn match_vec(plines: &[&str], s: &str) -> bool {
+pub(crate) fn match_vec(plines: &Vec<String>, s: &str) -> bool {
```
jacob-hughes: I'm confused here. Why do we need to pass a reference to the `Vec` instead? As far as I can tell, `match_vec` is only ever called on the same thread as it's caller, so you wouldn't need to use an `Arc` if you were passing a reference.

src/tester.rs
```diff
@@ -14,23 +14,34 @@ use std::{
...
-    test_file_filter: Option<Box<Fn(&Path) -> bool>>,
-    test_extract: Option<Box<Fn(&str) -> Option<String>>>,
-    test_cmds: Option<Box<Fn(&Path) -> Vec<(&str, Command)>>>,
     use_cmdline_args: bool,
+    test_file_filter: Option<Box<Fn(&Path) -> bool>>,
     cmdline_filters: Option<Vec<String>>,
+    inner: Arc<LangTesterInner>,
 }
 
+struct LangTesterInner {
```
jacob-hughes: It looks like you're splitting this struct so that you can share it between threads using only a single `Arc`. Perhaps `LangTesterShared` or `...SharedInner` (Although that's a bit long-winded) would make it's intention clearer.

src/fuzzy.rs
ltratt: This should be `&[String]` -- I'll fix that. But I think your question is really "why `String` and not `&str`?" I needed that in an intermediate transformation but I'm no longer sure if I need it. Let me try going back.

jacob-hughes: Okay

src/tester.rs
ltratt: I agree: the name I currently have sucks. `LangTesterShared` is better, though still not perfect. Suggestions welcome!

src/fuzzy.rs
ltratt: It is no longer needed, which is nice! The new commit fixes this, moving us back to `&str` and reducing copying.

src/tester.rs
ltratt: `LangTesterPoolable`?

jacob-hughes: Yes that's better.

ltratt:

jacob-hughes:
Looks good to me. As soon as the build passes please squash.

ltratt:
Squashed.

jacob-hughes:
Just noticed a typo in 537e362 "parallelise the readin".

ltratt:
OK, I'll have to force push to fix that. OK?

jacob-hughes:
Sure, that's fine.

ltratt:
Force pushed and updated.

PR #10, opened 2019-05-23
Remove "unsafe impl Sync" because we can make the user specify Sync.

ltratt:
This is a user visible API change, although I doubt few users would ever see the same. It's definitely worth us doing, as it means that we can't accidentally stumble into undefined behaviour if the user does have a non-Syncable function.

Fix from Niko Matsakis.


jacob-hughes:
Ouch yes. I hadn't thought of that.

PR #11, opened 2019-05-27
Print out all parts of a test which failed or which weren't checked.

ltratt:
Consider this test:

```
  VM:
    status: success
```

and assume it failed. Previously all we would print is the part of the test that failed i.e. the status check: all the user would see is "Error". This is correct, but not entirely helpful: did stderr/stdout contain useful clues?

This commit changes things so that any parts of a test which failed, or weren't referenced, will be printed out. In the example above, that means that stderr and stdout will also be printed out. But it also mean that if you have a test like:

```
  VM:
    status: success
    stdout:
```

and the test fails, but with nothing printed out to stdout (i.e. in this case, that means that the stdout test passed), then only the status and stderr will be printed out. In other words, we try and minimise the information printed to the user, while still giving them everything they might need to debug the problem.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -414,6 +414,9 @@ fn run_tests(
                 stdout: None,
             };
             for (cmd_name, mut cmd) in cmd_pairs {
+                let mut pass_status = None;
+                let mut pass_stderr = None;
+                let mut pass_stdout = None;
                 let output = cmd
                     .output()
                     .unwrap_or_else(|_| fatal(&format!("Couldn't run command {:?}.", cmd)));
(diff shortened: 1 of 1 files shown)


src/tester.rs
```diff
@@ -422,42 +425,62 @@ fn run_tests(
...
+                if let Some(ref stderr) = test.stderr {
+                    pass_stderr = Some(fuzzy::match_vec(stderr, &stderr_utf8));
+                }
+                let stdout_utf8 = String::from_utf8(output.stdout).unwrap();
+                if let Some(ref stdout) = test.stdout {
+                    pass_stdout = Some(fuzzy::match_vec(stdout, &stdout_utf8));
+                }
+
+                // If something failed, we want to print out everything which either failed a test
+                // or wasn't part of a test.
```
vext01: I'm not sure what you mean by printing stuff which isn't part of a test here.

ltratt: Does the commit message make this clearer?

vext01: I think I might have it. It prints anything that wasn't matched?

src/tester.rs
```diff
@@ -422,42 +425,62 @@ fn run_tests(
                     None => continue,
                 };
                 let mut meant_to_error = false;
+                // First, see which of the tests passed.
```
vext01: If I understand correctly, each execution of this closure runs one test?

```
for p in test_files {
    ...
    pool.execute(...
```

If so, I'm a little confused by this comment. Don't we know which test passed/failed due to the construction of the test loop?

Maybe I've misunderstood?

vext01: Ohhhh. You mean find out *how* it failed (out of {exit status, stderr match, stdout match}), not *which* test failed. I think we can tweak this comment to make it clearer.

src/tester.rs

src/tester.rs
ltratt: OK will tweak.

ltratt:

src/tester.rs
```diff
@@ -445,8 +446,8 @@ fn run_tests(
                     pass_stdout = Some(fuzzy::match_vec(stdout, &stdout_utf8));
                 }
 
-                // If something failed, we want to print out everything which either failed a test
-                // or wasn't part of a test.
+                // Second, if a test failed, we want to print out everything which didn't match
+                // (i.e. it either failed a test or wasn't part of a test).
```
vext01: At the risk of sounding really anal, the wording "wasn't part of a test" is the part I found confusing, and it's cropped up again here in brackets. I'd prefer something like "it either caused the test to fail, or was not used in the matching process due to an earlier mismatch".

vext01: How about:
"Second, if a test failed, we want to print out only the information releveant to the failure. For example, if the test doesn't specify matching rules for stderr, then don't print stderr".

ltratt: See if the new commit works for you.

vext01:
Please squash

ltratt:
Squashed.

PR #12, opened 2019-05-27
Exit the process on a fatal error.

ltratt:
The problem with running things multi-threaded is that a panic in one thread doesn't kill the others, so the user can easily end up not noticing that something very bad has happened. This commit turns panics into process-exiting fatal errors: when something bad happens, it's now very hard not to notice it!

### DIFF
--- a/src/fuzzy.rs
+++ b/src/fuzzy.rs
@@ -7,6 +7,8 @@
 // at your option. This file may not be copied, modified, or distributed except according to those
 // terms.
 
+use crate::fatal;
+
 const WILDCARD: &str = "...";
 
 /// Does `s` conform to the fuzzy pattern `pattern`? Note that `plines` is expected not to start or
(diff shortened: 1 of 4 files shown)


vext01:
If I understand correctly, this stops the tests on first error.

This is a bit more work, but wouldn't it be better to have each thread return an error status, allow the threads to join, and at the end print something like "there were failures, see above"? Then the rest of the tests still run.

ltratt:
Not exactly: this stops the tests on the first *internal* error. In other words, if the runner borked (or if the user used the wrong syntax in a test) it stops dead. "Normal" failures (i.e. a test failed) work exactly as before (i.e. like `cargo test`) and are batched up.

vext01:
I see. I'm OK with that then.

Let me review this in a little more detail now I understand the idea.

vext01:
Hrm, maybe if the user gives the wrong syntax, that shouldn't be fatal? The precedent in `compiletest` is to print a warning and continue, but I don't feel all that strongly tbh.

There are tests in rustc where they forgot to add any lines to match, so I see this fairly regularly.

ltratt:
For a future PR, I'm fine with making those things non-fatal, but that's really a job for a future PR (the plumbing won't be entirely trivial). This PR makes the current situation (where there's a `panic` of a thread which gets lost in a spew of normal output) less bad.

vext01:
OK. I'm OK with this then. Ready to merge?

ltratt:
Yep!

PR #13, opened 2019-05-29
Prepare a 0.2.1 release.

ltratt:


ltratt:
I've just realised/remembered that we've changed the API. This needs to be 0.3.0. Will reraise in a minute.

PR #15, opened 2019-06-02
Add support for altering the number of test threads

ltratt:
This is modelled after the same functionality in `cargo test`.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -36,6 +36,7 @@ pub struct LangTester<'a> {
 /// This is the information shared across test threads and which needs to be hidden behind an
 /// `Arc`.
 struct LangTesterPooler {
+    test_threads: usize,
     test_extract: Option<Box<Fn(&str) -> Option<String> + Send + Sync>>,
     test_cmds: Option<Box<Fn(&Path) -> Vec<(&str, Command)> + Send + Sync>>,
 }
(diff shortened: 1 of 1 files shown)


vext01:
I think I understand why the second commit exists, but this is not the behaviour of existing test suites. How hard is it to always have the option to print the test name regardless of thread count?

vext01:
Otherwise this all looks fine :)

ltratt:
> How hard is it to always have the option to print the test name regardless of thread count?

It is always printed: it's just a question of when.

In a single-threaded context, we print out the name *before* we execute the test file.

In a multi-threaded context, we print out the name *after* we execute the test file (otherwise the test file and "OK"/"Failed" would get separated). AFAICS this is what `cargo test` does too.

vext01:
> In a multi-threaded context, we print out the name after we execute the test file (otherwise the test file and "OK"/"Failed" would get separated).

OK, that's smart. Thanks for clarification.

vext01:
I'm good with this then. Perhaps alter the second commit message to include your clarification on when things are printed?

ltratt:
I can do that at the same time as squashing if you want?

vext01:
Sure. I wasn't sure if you'd squash or not. As you wish :)

ltratt:
Oops, I won't squash... but I will reword!

ltratt:
Reworded.

vext01:
OK

PR #16, opened 2019-06-02
Bail out as soon as a test has failed.

ltratt:
If we have several tests:

```
  A:
    stdout: ...

  B:
    stdout: ...
```

then if test A failed, we want to stop immediately, otherwise test B might still succeed and lead to confusion. In other words, stop on the first error, and print the output of that error to the user.

Most of this PR is a couple of comments to clarify the situation, but the newly added `break` is critical!

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -483,8 +483,13 @@ fn run_tests(
                     if pass_stdout.is_none() || pass_stdout == Some(false) {
                         failure.stdout = Some(stdout_utf8);
                     }
+
+                    // If a sub-test failed, bail out immediately, otherwise subsequent sub-tests
+                    // will overwrite the failure output!
+                    break;
                 }
 
+                // If a command failed, and we weren't expecting it to, bail out immediately.
                 if !output.status.success() && meant_to_error {
                     break;
                 }


vext01:
Are you sure this is what we want? Most test suites (cargo and pytest included) will run all tests and provide a failure report at the end. I think their strategy is well suited to CI, where you don't want to have to re-run lengthy build/test cycles to discover if more tests will fail.

ltratt:
For a file `x`, this just stops running its tests if `A` fails; file `b` (etc. etc.) is entirely independent.

vext01:
So we don't run further sub-tests within the same test? Is that correct?

ltratt:
Correct.

src/tester.rs
```diff
@@ -483,8 +483,13 @@ fn run_tests(
                     if pass_stdout.is_none() || pass_stdout == Some(false) {
                         failure.stdout = Some(stdout_utf8);
                     }
+
+                    // If a test failed, bail out immediately, otherwise subsequent tests will
```
vext01: So should we clarify that we are talking about sub-tests, not entire tests?

ltratt: Agreed.

ltratt:
Try that new commit.

vext01:
Thanks, that looks good.

Ready to squash?

ltratt:
Squashed (and reworded the message a bit, along the lines of the updated comment).

vext01:
This would be super-pedantry, but the commit message has potentially ambiguous references to "tests". Up to you if you think it's worth fixing.

ltratt:
Honestly, the terminology here is a mess. At some point I need to really think this through, but at the moment there is a fundamental ambiguity, and I think we have to live with it for a bit.

vext01:
OK. Hopefully we can revisit this later. For now, merging this.

PR #18, opened 2019-06-03
Get the terminology into some sort of order.

ltratt:
This was previously a mess, with the term "test" heavily overloaded. This commit tidies that up somewhat. Test files contain test data. Test data contains test commands. Test commands contain sub-tests. Perfect? No -- but definitely better than what came before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,9 @@ fn main() {
     let tempdir = TempDir::new("rust_lang_tester").unwrap();
     LangTester::new()
         .test_dir("examples/rust_lang_tester/lang_tests")
-        // Only use files named `*.rs` as tests.
+        // Only use files named `*.rs` as test files.
         .test_file_filter(|p| p.extension().unwrap().to_str().unwrap() == "rs")
-        // Extract the first sequence of commented line(s) as the test.
+        // Extract the first sequence of commented line(s) as the tests.
         .test_extract(|s| {
             Some(
                 s.lines()
@@ -58,10 +58,10 @@ fn main() {
 ```
 
 This defines a lang tester that uses all `*.rs` files in a given directory as
-tests, running two commands against them: `Compiler` (i.e. `rustc`); and
-`Run-time` (the compiled binary).
+test files, running two test commands against them: `Compiler` (i.e. `rustc`);
+and `Run-time` (the compiled binary).
 
-Users can then write files with tests and their inputs such as the following:
+Users can then write test files such as the following:
 
 ```rust
 // Compiler:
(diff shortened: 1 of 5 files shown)


vext01:
Let's see if I get this:

```
// Compiler:
//   status: success
//   stderr:
//     warning: unused variable: `x`
//       ...unused_var.rs:12:9
//       ...
//
// Run-time:
//   status: success
//   stdout: Hello world
...
```

* The whole file is a test file (duh) and the whole contents is the test data (are they the same then?)
* Looking at the changes,  `Compiler` and `Run-time` are "commands", but why that name? Because you ran a command to get the output?
* Sub-tests must be `status` and `stderr`? Would "test outputs" be better? stderr/stdout "output" text, whereas status "outputs" an integer? Too flimsy?

I agree that this is hard to name...

ltratt:
>    The whole file is a test file (duh) and the whole contents is the test data (are they the same then?)

The "test data" is whatever part of the test file contains test-esque thingies.

>    Looking at the changes, Compiler and Run-time are "commands", but why that name? Because you ran a command to get the output?

Yes.

>    Sub-tests must be status and stderr? Would "test outputs" be better? stderr/stdout "output" text, whereas status "outputs" an integer? Too flimsy?

`status` isn't really an output in the normal sense of the word :/

README.md
```diff
@@ -80,29 +80,29 @@ fn main() {
...
-content at all). `stderr`/`stdout` tests can use `...` as a simple wildcard: if
-a line consists solely of `...`, it means "match zero or more lines"; if a line
-begins with `...`, it means "match the remainder of the line only"; if a line
-ends with `...`, it means "match the start of the line only". A line may start
-and end with `...`. `stderr`/`stdout` matches ignore leading/trailing
-whitespace and newlines, but are case sensitive.
+Test data is specified with a two-level indentation syntax: the outer most
+level of indentation defines a test command (multiple command names can be
+specified, as in the above); each test command can then define sub-tests for
+one or more of `status: <success|failure|<int>>` (where `success` and `failure`
```
vext01: I wonder if we should bullet list this part for readability and to break up the text a bit?

README.md
```diff
@@ -80,29 +80,29 @@ fn main() {
...
+
+The above file thus contains 4 actual tests: the `Compiler` should succeed
+(e.g. return a `0` exit code when run on Unix), and its `stderr` output should
+warn about an unused variable on line 12; and the resulting binary should
+succeed and produce `Hello world` on `stdout`.
+
+Potential sub-tests that are not mentioned are not tested: for example, the
+above file does not state whether the `Compiler`s `stdout` should have content
+or not (but note that the line `stdout:` on its own would state that the
+`Compiler` should have no content at all).  `stderr`/`stdout` sub-tests can use
```
vext01: And I think we should starta  new paragraph at `stderr/stdout` here?

vext01: And any changes here need to go into the doc comments too.

It's a shame it has to be duplicated.

ltratt: Even worse than being duplicated, it's got different line prefixes and line widths ;)

ltratt:

ltratt:
And try that again (sorry!).

PR #21, opened 2019-07-30
Really delete blank lines.

ltratt:
We stated this as an invariant for fuzzy::match_vec but the code that should have enforced it didn't work properly. Mea culpa!

As well as fixing this (with, perhaps inevitably, simpler code than the buggy code it replaces!), this commit adds asserts to stop this ever happening again.


ltratt:
I just realised that I could add a couple of simple tests to make clear the problem which I've just pushed: that commit should probably be squashed away.

vext01:
Please squash.

ltratt:
Squashed.

PR #25, opened 2019-10-22
Add ability to pass per test command line args

jacob-hughes:
Sometimes it's desirable to run certain tests (or subcommands) with custom command line arguments. This commit adds the ability to do this without the need for defining a separate test harness.

We introduce a new keyword, `args`, which can be optionally specified for each test command. For example, suppose an individual test relied on handling some user input, one could pass this input as follows:

```
Run-time:
    args: 1234
    status: success
```

This passes the command line argument 1234 to the command named "Run-time". Note that command line arguments are always passed as strings.

It is important to remember that custom arguments should not break invariants which are relied upon by the test runner. For example, suppose a test runner has two commands: compile, and then execute a C program. If an individual test provides a custom argument to specify a different output for the compiled binary:

```
Compile:
    args: -o /tmp/
    ...
```

Then the test harness will fail, as it will not know to look in the optionally specified /tmp/ directory.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -349,6 +349,9 @@ pub(crate) struct TestCmd<'a> {
     pub status: Option<Status>,
     pub stderr: Option<Vec<&'a str>>,
     pub stdout: Option<Vec<&'a str>>,
+    /// A list of custom command line arguments which should be passed when
+    /// executing the test command.
+    pub args: Vec<String>,
 }
 
 /// If one or more parts of a `TestCmd` fail, the parts that fail are set to `Some(...)` in an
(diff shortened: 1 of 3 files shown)


src/tester.rs
```diff
@@ -349,6 +349,9 @@ pub(crate) struct TestCmd<'a> {
     pub status: Option<Status>,
     pub stderr: Option<Vec<&'a str>>,
     pub stdout: Option<Vec<&'a str>>,
+    /// A list of custom command line arguments which should be passed when
+    /// executing the test command.
+    pub args: Option<Vec<String>>,
```
ltratt: Since (under the hood) a `Vec` doesn't allocate anything until you push to it, it would probably be easier to make this `Vec<String>`. I think that will simplify some other code?

ltratt:
I like this idea!

src/tester.rs
jacob-hughes: Yeah that makes sense. See what you think now.

ltratt:
Please squash.

jacob-hughes:
Squashed

ltratt:
I have just had a thought. This *adds* args to the list passed to the command, right? So maybe we should call the flag `extraargs` (or similar) so that people don't think it's overriding args? [And, if anyone has a valid use case for it in the future, I suppose we could add `overrideargs` as a separate thing.]

jacob-hughes:
Yes, I think you're right, these are additional arguments and we should be careful to word this in a way that doesn't imply that it overrides the args.

I've renamed the keyword `extra-args`, as double vowel in  `extraargs` irritates me more than it should.

ltratt:
Bah, I hate `-` but it's your PR :) Please squash.

jacob-hughes:
I would have suggested underscore but it clashes with `Run-time` ;)

jacob-hughes:
Squashed

PR #29, opened 2019-10-25
Explicitly use defaults

ltratt:
This PR embraces the new era of lang_tester using defaults to simplify the code base. The motivation is contained in the doc changes of https://github.com/softdevteam/lang_tester/commit/e13d06647235906f28e92a05dce8c2acf674fa3b, with the code then being easily simplified (in a way that's semantically backwards compatible, though note that it does change the output shown to users when tests fail) in https://github.com/softdevteam/lang_tester/commit/dc4e11c14926dffdc790afa5253d5a26d417791a. This PR deletes more code than it adds! Please test, though, on your own code to check that it doesn't do anything silly!

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -346,20 +346,20 @@ pub(crate) enum Status {
 /// A user `TestCmd`.
 #[derive(Clone, Debug)]
 pub(crate) struct TestCmd<'a> {
-    pub status: Option<Status>,
-    pub stderr: Option<Vec<&'a str>>,
-    pub stdout: Option<Vec<&'a str>>,
+    pub status: Status,
+    pub stderr: Vec<&'a str>,
+    pub stdout: Vec<&'a str>,
     /// A list of custom command line arguments which should be passed when
     /// executing the test command.
     pub args: Vec<String>,
 }
 
 impl<'a> TestCmd<'a> {
-    fn default() -> Self {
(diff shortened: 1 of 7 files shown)


src/tester.rs
```diff
@@ -346,20 +346,20 @@ pub(crate) enum Status {
...
 
 impl<'a> TestCmd<'a> {
-    fn default() -> Self {
+    pub fn default() -> Self {
         Self {
-            status: Some(Status::Success),
-            stderr: None,
-            stdout: None,
+            status: Status::Success,
+            stderr: vec!["..."],
```
jacob-hughes: Why not an empty `Vec` here?

ltratt: That would only match the empty string i.e. it would mean that a command that produced any output wouldn't match successfully. I think the more permissive `...` is preferable myself (i.e. match any output, including none at all).

jacob-hughes: Ah yes that makes sense. I agree, by default we should allow any output.

jacob-hughes:
LGTM. Much cleaner, I like delete PRs :)

I've tested this on `gcmalloc` and it runs as expected.

jacob-hughes:
I'm guessing if you're happy, I can merge as is?

ltratt:
Works for me!

PR #30, opened 2019-10-25
Add support for --nocapture

ltratt:
This adds support for `--nocapture` to lang_tester. The main part of the PR is https://github.com/softdevteam/lang_tester/commit/d0ac369a66d84c3f246c1e7c8d0706456a1b1023. As with normal `cargo test`, `--nocapture` on its own tends to lead to baffling interleaving of commands: it generally pairs well with `--test-threads=1`.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -11,22 +11,29 @@ use std::{
     collections::{hash_map::HashMap, HashSet},
     env,
     fs::read_to_string,
-    io::{self, Write},
+    io::{self, Read, Write},
     path::{Path, PathBuf},
     process::{self, Command, ExitStatus},
+    str,
     sync::{Arc, Mutex},
-    time::Duration,
+    time::{Duration, Instant},
 };
 
+use filedescriptor::{
+    poll, pollfd, FileDescriptor, IntoRawSocketDescriptor, POLLERR, POLLHUP, POLLIN,
+};
 use getopts::Options;
 use num_cpus;
 use termcolor::{Color, ColorChoice, ColorSpec, StandardStream, WriteColor};
 use threadpool::ThreadPool;
-use wait_timeout::ChildExt;
 use walkdir::WalkDir;
 
 use crate::{fatal, fuzzy, parser::parse_tests};
 
+/// The size of the (stack allocated) buffer use to read stderr/stdout from a child process.
+const READBUF: usize = 1024 * 4; // bytes
+/// Print a warning to the user every multiple of `TIMEOUT` seconds that a child process has run
+/// without completing.
 const TIMEOUT: u64 = 60; // seconds
 
 pub struct LangTester<'a> {
@@ -41,6 +48,7 @@ pub struct LangTester<'a> {
 /// `Arc`.
 struct LangTesterPooler {
     test_threads: usize,
+    nocapture: bool,
     test_extract: Option<Box<dyn Fn(&str) -> Option<String> + Send + Sync>>,
     test_cmds: Option<Box<dyn Fn(&Path) -> Vec<(&str, Command)> + Send + Sync>>,
 }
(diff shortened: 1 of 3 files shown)


ltratt:
https://github.com/softdevteam/lang_tester/pull/30/commits/45b532ca291403e8cd165657da9dc580954dbd3f will need squashing away, but fixes a mistake I introduced in https://github.com/softdevteam/lang_tester/pull/30/commits/d0ac369a66d84c3f246c1e7c8d0706456a1b1023.

src/tester.rs
```diff
@@ -569,30 +583,103 @@ fn run_cmd(
...
+                        break;
+                    }
+                    cap_stdout.extend_from_slice(&buf[..i]);
+                    if inner.nocapture {
+                        print!("{}", str::from_utf8(&buf).unwrap());
+                    }
+                }
+            }
+
+            if pollfds[0].revents & POLLHUP == POLLHUP && pollfds[1].revents & POLLHUP == POLLHUP {
```
jacob-hughes: If `stdout` and `stdin` have been disconnected?

src/tester.rs
```diff
@@ -569,30 +583,103 @@ fn run_cmd(
...
+            next_warning = last_warning
+                .checked_add(Duration::from_secs(TIMEOUT))
+                .unwrap();
+        }
     }
-    let output = child.wait_with_output().unwrap();
-    (output.status, output.stderr, output.stdout)
+
+    let status = child
+        .wait()
```
jacob-hughes: Should we `wait_timeout` here, in case the test doesn't terminate for whatever reason?

src/tester.rs
```diff
@@ -569,30 +583,103 @@ fn run_cmd(
...
+            let t = Instant::now();
+            if t > next_warning {
+                Duration::from_secs(1)
+            } else {
+                next_warning.duration_since(t)
             }
         };
-        looped += 1;
+        if let Ok(_) = poll(&mut pollfds, Some(timeout)) {
+            if pollfds[0].revents & POLLIN == POLLIN {
```
jacob-hughes: It looks like we only check the first file descriptor in `pollfds`, which is `stderr`, when does  `stdout` get checked?

src/tester.rs

src/tester.rs
```diff
@@ -569,30 +583,103 @@ fn run_cmd(
...
+
     let mut child = cmd
         .stderr(process::Stdio::piped())
         .stdout(process::Stdio::piped())
         .spawn()
         .unwrap_or_else(|_| fatal(&format!("Couldn't run command {:?}.", cmd)));
 
-    let mut looped = 1;
+    let mut stderr = FileDescriptor::dup(child.stderr.as_ref().unwrap()).unwrap();
+    let mut stdout = FileDescriptor::dup(child.stdout.as_ref().unwrap()).unwrap();
```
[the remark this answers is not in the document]
ltratt: `stderr` and `stdout` aren't always guaranteed to be those values AFAIK. To be honest, this seems a bit of a dance, but I more-or-less followed what the `FileDescriptor` crate suggested.

src/tester.rs
ltratt: Good idea!

src/tester.rs
ltratt: Line 649(ish) handles `stdout`.

jacob-hughes: Ah yes, not sure why I missed this!

src/tester.rs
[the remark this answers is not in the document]
jacob-hughes: I understand. That's fine with me

ltratt:
@jacob-hughes https://github.com/softdevteam/lang_tester/pull/30/commits/581e2e006233bb36a6ce41aa4d240363c181c7ef implements "wait after stderr/stdout have been closed". There's no native API for this so we have to roll our own.

https://github.com/softdevteam/lang_tester/pull/30/commits/dc23e6e7ac41636cf9460e390e9f4db4a934c152 and https://github.com/softdevteam/lang_tester/pull/30/commits/b0f991f52f878d5fa0b2b904388873029d1bdabe are two minor changes I noticed are probably worth doing while I'm here. The latter should be squashed later IMHO.

jacob-hughes:
LGTM. Please squash.

ltratt:
Squashed.

PR #31, opened 2019-10-28
Prepare a 0.3.4 release.

ltratt:


jacob-hughes:
As this is a new feature, I think this should be a minor bump instead of a patch.

jacob-hughes:
I've just noticed the date is wrong. Please force push an update.

ltratt:
Updated.

PR #33, opened 2019-11-03
Remmove quotes from output.

ltratt:
The extra `'` characters had crept in during testing and I didn't notice them when making a previous PR. Mea culpa!


jacob-hughes:
I had assumed this was intentional! :)

PR #35, opened 2019-11-12
Show SIGNUM instead of Error if test fails due to thrown signal

jacob-hughes:
When a program terminates due to thrown signal, it's usually helpful for
debugging to know why. For example, if `lang_tester` runs a command
which we expect to succeed and instead segfaults, `lang_tester` will now
output (on linux):

"Exited due to signal: 11"

Instead of:

"Error"

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -12,6 +12,7 @@ use std::{
     env,
     fs::read_to_string,
     io::{self, Read, Write},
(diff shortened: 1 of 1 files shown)


jacob-hughes:
I've removed the failing test now.

ltratt:
Please squash.

jacob-hughes:
Squashed.

PR #36, opened 2019-11-12
Add status for specifying that a signal was caught

jacob-hughes:
If the user expects a command, `x`, to terminate due to a signal being
thrown, they can now specify this explicitly, rather than using the
generic "error" status.

A test will still pass if it expects an "error" status from `x` which
terminated with a signal, as "error" is a superset of "signal".

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -354,7 +354,7 @@ impl<'a> LangTester<'a> {
 }
 
 /// The status of an executed command.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) enum Status {
     /// The command exited successfully (by whatever definition of "successful" the running
     /// platform uses).
@@ -408,6 +408,24 @@ fn write_with_colour(s: &str, colour: Color) {
     stderr.reset().ok();
 }
 
+fn write_ignored(test_name: &str, message: &str, inner: Arc<LangTesterPooler>) {
+    // Grab a lock on stderr so that we can avoid the possibility of lines blurring
+    // together in confusing ways.
+    let stderr = StandardStream::stderr(ColorChoice::Always);
+    let mut handle = stderr.lock();
+    if inner.test_threads > 1 {
+        handle
+            .write_all(&format!("\ntest lang_tests::{} ... ", test_name).as_bytes())
+            .ok();
+    }
+    handle
+        .set_color(ColorSpec::new().set_fg(Some(Color::Yellow)))
+        .ok();
+    handle.write_all(b"ignored").ok();
+    handle.reset().ok();
+    handle.write_all(format!(" ({})", message).as_bytes()).ok();
+}
+
 fn usage() -> ! {
     eprintln!("Usage: [--test-threads=<n>] <filter1> [... <filtern>]");
     process::exit(1);
@@ -453,27 +471,30 @@ fn run_tests(
             let test_str = inner.test_extract.as_ref().unwrap()(&all_str).unwrap_or_else(|| {
                 fatal(&format!("Couldn't extract test string from {}", test_fname))
             });
+
             if test_str.is_empty() {
-                // Grab a lock on stderr so that we can avoid the possibility of lines blurring
-                // together in confusing ways.
-                let stderr = StandardStream::stderr(ColorChoice::Always);
-                let mut handle = stderr.lock();
-                if inner.test_threads > 1 {
-                    handle
-                        .write_all(&format!("\ntest lang_tests::{} ... ", test_fname).as_bytes())
-                        .ok();
-                }
-                handle
-                    .set_color(ColorSpec::new().set_fg(Some(Color::Yellow)))
-                    .ok();
-                handle.write_all(b"ignored").ok();
-                handle.reset().ok();
-                handle.write_all(b" (test string is empty)").ok();
+                write_ignored(test_fname.as_str(), "test string is empty", inner);
                 num_ignored += 1;
                 return;
             }
 
             let tests = parse_tests(&test_str);
+
+            if !cfg!(unix)
+                && tests
+                    .values()
+                    .find(|t| t.status == Status::Signal)
+                    .is_some()
+            {
+                write_ignored(
+                    test_fname.as_str(),
+                    "signal termination not supported on this platform",
+                    inner,
+                );
+                num_ignored += 1;
+                return;
+            }
+
             let cmd_pairs = inner.test_cmds.as_ref().unwrap()(p.as_path())
                 .into_iter()
                 .map(|(test_name, cmd)| (test_name.to_lowercase(), cmd))
@@ -529,13 +550,7 @@ fn run_tests(
                                 }
                             }
                             Status::Signal => {
-                                if cfg!(unix) {
-                                    failure.status = Some("Did not exit due to signal".to_owned());
-                                } else {
-                                    failure.status = Some(
-                                        "Use of signal not supported on this platform".to_owned(),
-                                    )
-                                }
+                                failure.status = Some("Exit was not due to signal".to_owned());
                             }
                             Status::Int(_) => {
                                 failure.status = Some(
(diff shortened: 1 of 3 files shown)


src/tester.rs
```diff
@@ -524,6 +528,15 @@ fn run_tests(
                                     failure.status = Some("Error".to_owned());
                                 }
                             }
+                            Status::Signal => {
+                                if cfg!(unix) {
+                                    failure.status = Some("Did not exit due to signal".to_owned());
+                                } else {
```
ltratt: I think that in this case we can probably sensibly treat `Status::Signal` the same as `Status::Error`?

ltratt:
Once we've worked out exactly what we want this to do (and I don't think that'll take long), we'll also need to update the docs in `README.md` and in `lib.rs` (yes, they're near duplicates, which is annoying!).

src/tester.rs
jacob-hughes: I can see compelling arguments for both sides really. Ultimately, something that terminates with a signal in UNIX will probably also fail in windows with whatever their termination mechanism is. For the sake of compatibility with other platforms I think I agree with you.

Similarly, if we go down this route, we should probably aim for the same fallback-to-error behaviour with `Status::Int`.

ltratt: Maybe we should make `Status::Signal` an entirely compile-time switch so that on non-Unix platforms it's not even compiled in?

jacob-hughes: What would the behaviour be if I ran a test with `status: signal` on a windows machine? A test ignored message in the runner output?

ltratt: I think in the parser we'd just convert `status: signal` to `Status::Int`.

jacob-hughes: You mean to `Status::Error`?

ltratt: Mea culpa! Yes you're right.

jacob-hughes: Hmm, I don't think that's really much different to the previous option. I think if we go to the effort of conditionally compiling `Status::Signal` related code, we should only run tests that rely on it on unix platforms.

ltratt: OK, well we do have support for ignoring tests, so I guess we could do that on non-Unix. I'm easy either way.

jacob-hughes: I prefer that way. I can imagine a scenario where a program which terminates with a signal on Unix is valid on Windows. In which case, falling back to Status::Error would not be the correct behaviour. I'll push a commit that changes this.

jacob-hughes:
I've updated the behaviour of `lang_tester` to ignore tests for `Status::Signal` on non-Unix platforms. Let me know what you think.

src/tester.rs
```diff
@@ -445,35 +463,30 @@ fn run_tests(
         let failures = failures.clone();
         let inner = inner.clone();
         pool.execute(move || {
-            if inner.test_threads == 1 {
+            let single_threaded = inner.test_threads == 1;
```
ltratt: I'm not sure this assignment is useful given that it's only read from once.

src/tester.rs
```diff
@@ -431,6 +431,24 @@ fn check_names<'a>(cmd_pairs: &[(String, Command)], tests: &HashMap<String, Test
     }
 }
 
+fn write_ignored(test_name: &str, message: &str, single_threaded : bool) {
```
ltratt: Can we move this function to just after (or before) `write_with_colour`?

src/tester.rs
jacob-hughes: It's passed into `write_ignored` because the output differs depending on whether or not we're in a single-threaded context.

src/tester.rs
```diff
@@ -431,6 +431,24 @@ fn check_names<'a>(cmd_pairs: &[(String, Command)], tests: &HashMap<String, Test
...
+    let mut handle = stderr.lock();
+    if !single_threaded {
+        handle
+            .write_all(&format!("\ntest lang_tests::{} ... ", test_name).as_bytes())
+            .ok();
+    }
+    handle
+        .set_color(ColorSpec::new().set_fg(Some(Color::Yellow)))
+        .ok();
+    handle.write_all(b"ignored").ok();
```
ltratt: I know I was responsible for the original code here, but I think we can now simplify lines 447-449 to:

```rust
handle.write_all(format!("ignored ({})").as_bytes()).ok();
```

?

src/tester.rs
```diff
@@ -445,35 +463,30 @@ fn run_tests(
...
-                handle.write_all(b" (test string is empty)").ok();
+                write_ignored(test_fname.as_str(), "test string is empty", single_threaded);
                 num_ignored += 1;
                 return;
             }
 
             let tests = parse_tests(&test_str);
+
+            if !cfg!(unix) && tests.values().find(|t| t.status == Status::Signal).is_some() {
+                write_ignored(test_fname.as_str(), "platform not supported", single_threaded);
```
ltratt: Maybe we can say something like "signal exits not supported on this platform"? It is a slight mouthful, I admit, so if we can be terser, I'm all ears!

src/tester.rs
```diff
@@ -551,13 +551,7 @@ fn run_tests(
...
                             }
                             Status::Signal => {
-                                if cfg!(unix) {
-                                    failure.status = Some("Did not exit due to signal".to_owned());
-                                } else {
-                                    failure.status = Some(
-                                        "Use of signal not supported on this platform".to_owned(),
-                                    )
-                                }
+                                failure.status = Some("Did not exit due to signal".to_owned());
```
ltratt: This could be parsed as "did not exit all, signal or not". I think we mean something like "Exit was not due to signal"?

src/tester.rs
ltratt: Ah, I see! I think it might fit more with the other code to pass `inner: Arc<LangTesterPooler>` as an argument rather than `single_threaded`?

ltratt:
I have various minor comments, but the most important thing is that this looks like the right thing overall.

src/tester.rs
jacob-hughes: We do this because `ignored` should be orange and the text inside should be the default output colour. If we simplify it to that one-liner and don't reset the colour after then it bleeds onto the next line.

src/tester.rs
jacob-hughes: I can't think of anything better really. Lets go with this.

src/tester.rs
ltratt: Aha, I knew my past self was not as stupid as my current self :p

ltratt: Aha, I knew my past self was not as stupid as my current self :p

src/tester.rs
jacob-hughes: Fixed: 91772b5b2346a02919a49eabbac9f10c5e9f8c90

src/tester.rs
jacob-hughes: Fixed: 86ff69501c2fb06148d612175836666cdebae040

src/tester.rs
jacob-hughes: Fixed: 86ff69501c2fb06148d612175836666cdebae040

src/tester.rs
jacob-hughes: Fixed: 46ce135cdb6444e9f261fc5077f0db01bca99255

src/tester.rs
ltratt: I think you might have referenced the wrong commit. I think you meant https://github.com/softdevteam/lang_tester/pull/36/commits/729678dd3cbcb414eaf6a49dd655a958d731fa15.

jacob-hughes: Well done you passed the test.

ltratt:
Before I ask you to squash can I ask: were you able to test this somehow (given that we all probably only have Unix systems)? Perhaps you just temporarily deleted the `cfg!(unix)` clause?

jacob-hughes:
Yes I inverted the `cfg(unix)` and ran the examples :)

ltratt:
Just checking :)

Please squash.

jacob-hughes:
Squashed

PR #39, opened 2019-11-20
Move to our new licensing regime.

ltratt:
All copyright in this project is held by King's, so moving it is trivial.


COPYRIGHT
```diff
@@ -0,0 +1,19 @@
+Except as otherwise noted (below and/or in individual files), this project is
+licensed under the Apache License, Version 2.0 <LICENSE-APACHE> or
```
smarr: there seems to be an `or` too much here.

ltratt: You're right -- me culpa!

PR #42, opened 2019-11-25
Ignorable tests

ltratt:
Sometimes we might want to add tests that we can't yet run successfully, so being able to ignore them can be useful.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -46,6 +46,7 @@ pub struct LangTester<'a> {
 /// `Arc`.
 struct LangTesterPooler {
     test_threads: usize,
+    ignored: bool,
     nocapture: bool,
     test_extract: Option<Box<dyn Fn(&str) -> Option<String> + Send + Sync>>,
     test_cmds: Option<Box<dyn Fn(&Path) -> Vec<(&str, Command)> + Send + Sync>>,
@@ -62,6 +63,7 @@ impl<'a> LangTester<'a> {
             use_cmdline_args: true,
             cmdline_filters: None,
             inner: Arc::new(LangTesterPooler {
+                ignored: false,
                 nocapture: false,
                 test_threads: num_cpus::get(),
                 test_extract: None,
@@ -253,6 +255,7 @@ impl<'a> LangTester<'a> {
             let args: Vec<String> = env::args().collect();
             let matches = Options::new()
                 .optflag("h", "help", "")
+                .optflag("", "ignored", "Run only ignored tests")
                 .optflag(
                     "",
                     "nocapture",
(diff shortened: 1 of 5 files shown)


jacob-hughes:
For familiarity with `rustc --test`, I think it might be useful for `lang_tester` to respect the `--ignored` flag. In other words, there should be a mechanism to explicitly run ignored tests, as some projects mark expensive tests as ignored so that they only run when requested. (looking at you `rustc` with your hour long bootstrap test :eyes: )

ltratt:
Good idea. I should probably do both:

```
        --include-ignored 
                        Run ignored and not ignored tests
        --ignored       Run only ignored tests
```

ltratt:
Hmm, `--include-ignored` is unstable so I'll ignore that for now.

jacob-hughes:
Ah, I wasn't aware of `--include-ignored`.

ltratt:
https://github.com/softdevteam/lang_tester/pull/42/commits/ee06254cb0b673de26e59c3329624a4a60bb93e4 adds support for `--ignored` -- but please test!

src/tester.rs
```diff
@@ -470,150 +484,163 @@ fn run_tests(
...
-                    .find(|t| t.status == Status::Signal)
-                    .is_some()
-            {
-                write_ignored(
-                    test_fname.as_str(),
-                    "signal termination not supported on this platform",
-                    inner,
-                );
+            if (inner.ignored && !tests.ignore) || (!inner.ignored && tests.ignore) {
+                write_ignored(test_fname.as_str(), "", inner);
```
jacob-hughes: Was there a reason not to pass the ignore string from top of the test string here? I think it would be nice but I don't have a very strong opinion on this.

src/tester.rs
```diff
@@ -470,150 +484,163 @@ fn run_tests(
...
             }
-        });
+
+            // If a sub-test failed, bail out immediately, otherwise subsequent sub-tests
+            // will overwrite the failure output!
+            break;
+        }
+
+        // If a command failed, and we weren't expecting it to, bail out immediately.
+        if !status.success() && meant_to_error {
```
jacob-hughes: Is this right? The comment seems to contradict the conditional.

jacob-hughes:
I've tested this on `gcmalloc` and both ignoring tests and the `--ignored` flag seem to work. LGTM other than a few minor comments.

src/tester.rs
ltratt: It's just a bit annoying to pass it through :/

src/tester.rs
ltratt: I think it obeys it? I didn't change this bit anyway (other than moving the code, which I admit makes it look like a change).

jacob-hughes: Fair enough. Lets leave it alone if there was no functional change!

src/tester.rs
jacob-hughes: I thought it might be something like that. That's fine it's not that important.

jacob-hughes:
Please squash

ltratt:
I've updated the commit message for the third / final commit.

jacob-hughes:
Did you mean to delete the last paragraph in ed3b0bee61f511bdfe1fd8e6a2a6694744e31e97?

ltratt:
Mea culpa! That shouldn't have been left in. I've forced pushed an update. Hopefully this is ready now!

PR #43, opened 2019-11-26
Clippy fixes and force Clippy to run in Travis

ltratt:
This PR makes lang_tester have zero Clippy warnings, so let's be brave and make sure that Clippy never complains again after a PR!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: rust
 before_script:
-    - rustup toolchain install nightly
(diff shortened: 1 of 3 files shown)


ltratt:
@jacob-hughes If you're OK with this, I need to do a bit of squashing!

jacob-hughes:
LGTM please squash

ltratt:
Force pushed (and slightly updated the last commit message).

Issue #46, opened 2019-12-01
PR #30 causes tests to never end on macOS.


ummarikar:
After testing this repository with https://github.com/softdevteam/yksom, I have found that #30, specifically [`#a8f8044`](https://github.com/softdevteam/lang_tester/pull/30/commits/a8f80443f0ec2bf983aa22ee479d725e91459488), has caused running `cargo test` to never end on macOS (tested on two Macs):
<img width="554" alt="Screenshot 2019-11-27 at 17 38 32" src="https://user-images.githubusercontent.com/32270615/69919910-e0014780-1479-11ea-9c93-c5f1b465eaec.png">

ltratt:
Can you propose a fix (preferably raising a PR to fix it)? I suspect that searching for "OS X and poll" might be a fruitful place to start!

ltratt:
This *might* be relevant but I'm not sure https://www.greenend.org.uk/rjk/tech/poll.html

ummarikar:
I'll see what I can do :)

ummarikar:
UPDATE: the poll command on OSX works. I found this out by running the C code at this link - https://daniel.haxx.se/blog/2016/10/11/poll-on-mac-10-12-is-broken/.  https://github.com/softdevteam/lang_tester/blob/a1f51ec8b747cb252a0b8df4846ff43c8c0c6c8b/src/tester.rs#L729-L731
This if statement is never entered (on OSX) as the condition is always false. Each side of the conjunction is always false whereas this if statement is always executed on Linux. I believe this is what is causing the loop.

ummarikar:
I found that the difference between Linux and MacOS using filedescriptor `poll` is the following:
```The filedescriptor crate offers a poll(2) compatible interface suitable for testing the readiness of a set of file descriptors. On unix systems this is a very thin wrapper around poll(2), except on macOS where it is actually a wrapper around the select(2) interface.```
Maybe this is relevant? Not too sure just speculating at the moment.

ummarikar:
> UPDATE: the poll command on OSX works. I found this out by running the C code at this link - https://daniel.haxx.se/blog/2016/10/11/poll-on-mac-10-12-is-broken/.
> 
> https://github.com/softdevteam/lang_tester/blob/a1f51ec8b747cb252a0b8df4846ff43c8c0c6c8b/src/tester.rs#L729-L731
> 
> 
[...]

PR #47, opened 2019-12-08
Fix for issue #46 - lang_tester runs into a loop on MacOS.

ummarikar:
This is an in-progress PR to show the steps taken to solve issue #46.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -15,9 +15,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use filedescriptor::{
-    poll, pollfd, FileDescriptor, POLLHUP, POLLIN, AsRawSocketDescriptor, IntoRawSocketDescriptor
-};
+use filedescriptor::{poll, pollfd, AsRawSocketDescriptor, FileDescriptor, POLLHUP, POLLIN};
 use getopts::Options;
 use num_cpus;
 use termcolor::{Color, ColorChoice, ColorSpec, StandardStream, WriteColor};
@@ -653,8 +651,8 @@ fn run_cmd(
         .spawn()
         .unwrap_or_else(|_| fatal(&format!("Couldn't run command {:?}.", cmd)));
 
- let mut stderr = FileDescriptor::dup(child.stderr.as_ref().unwrap()).unwrap(); 
- let mut stdout = FileDescriptor::dup(child.stdout.as_ref().unwrap()).unwrap();
+    let mut stderr = FileDescriptor::dup(child.stderr.as_ref().unwrap()).unwrap();
+    let mut stdout = FileDescriptor::dup(child.stdout.as_ref().unwrap()).unwrap();
     // Buffers to accumulate the child output
     let mut output = Vec::new();
     let mut error = Vec::new();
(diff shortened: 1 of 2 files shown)


ummarikar:
Screenshot 2019-12-08 at 23 53 23
This is the current error I'm getting my fix was to switch (back) to using the following lines:
https://github.com/softdevteam/lang_tester/blob/a1f51ec8b747cb252a0b8df4846ff43c8c0c6c8b/src/tester.rs#L661-L662
Pushing the changes now.

ummarikar:
This solves the error I was getting however now the tests take an incredibly long time to run and when they finally show up they fail. I had to cancel the run however as it was taking too long.

ummarikar:
In my local version of `filedescriptor` I switched from macOS's implementation of poll(2) which is a wrapper around select(2) to using the actual poll(2) on macOS (obviously this is not an option and was just done for experimentation purposes). The result was that the yksom tests actually run and don't take an incredibly long time but they all fail. And, it looks like the expected output is being added to stderr:
Screenshot 2019-12-09 at 00 22 17
Looking into this now.

ummarikar:
Tests now all pass when using the macOS `poll(2)` in `filedescriptor` but using the `select(2)` implementation produces the same problem I was getting before where the tests wouldn't run.

ltratt:
OK, let's break this PR down into some stages and start again (it will probably be easier to close this and open a new PR, but that's up to you):

  1) Use the suggestion from https://github.com/wez/wzsh/pull/1 for converting pipes into `FileDescriptor` objects. That's a commit on its own.

  2) The child stdin/stderr file descriptors need to be marked as non-blocking. I don't know if the `filedescriptor` library allows that. If not, you'll need to use `fcntl`. This should be a commit on its own. There is a tiny chance that this might worm around the `poll` bug on OS X, at which point job done! If not, well, it needed doing anyway.

  3) If `poll` still doesn't work, then use old fashioned `select` instead. This will require a bit more work.

ummarikar:
I'll keep using this PR then we can squash everything where appropriate when we're done :)

ummarikar:
Non-blocking implemented. I'm wondering if I have done this correctly. I want to be sure I have before I move on to trying the `select(2)` route.

src/tester.rs
```diff
@@ -651,8 +652,14 @@ fn run_cmd(
         .spawn()
         .unwrap_or_else(|_| fatal(&format!("Couldn't run command {:?}.", cmd)));
 
-    let mut stderr = FileDescriptor::dup(child.stderr.as_ref().unwrap()).unwrap();
     let mut stdout = FileDescriptor::dup(child.stdout.as_ref().unwrap()).unwrap();
+    let mut stderr = FileDescriptor::dup(child.stderr.as_ref().unwrap()).unwrap();
```
ltratt: I'm not sure why this line has moved.

src/tester.rs
```diff
@@ -651,8 +652,14 @@ fn run_cmd(
...
+
+    fcntl(stdout.as_socket_descriptor(), F_SETFL(OFlag::O_NONBLOCK))
```
ltratt: I'm pretty sure that you need to do `flags = fcntl(..., F_GETFL, 0)` to read the flags, then `fcntl(..., F_SETFL, flags | O_NONBLOCK)`.

PR #48, opened 2019-12-13
Implement non-blocking.

ummarikar:
Implement non-blocking for file descriptors. This works with `filedescriptor` using its `poll(2)` implementation but doesn't work with it using its `select(2)` implementation. However, I want to be utterly sure that I have implemented non-blocking properly before I try other things like implementing our own `select(2)` solution.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ path = "examples/rust_lang_tester/run_tests.rs"
 [dependencies]
 filedescriptor = "0.5"
 getopts = "0.2"
+nix = "0.16.0"
 num_cpus = "1.10"
 termcolor = "1.0"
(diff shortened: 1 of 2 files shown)


ltratt:
This has got too much stuff in. Let's please do as I suggested: 1 commit that *just* does non-blocking (i.e. `fcntl` -- unfortunately I commented on the old PR https://github.com/softdevteam/lang_tester/pull/47 but you can see the comments there). Then another commit which does other things related to the buffer stuff.

ummarikar:
Oh I see. I assumed that the buffers would be included in the first commit. Apologies for that. I shall send another commit and if you’re happy with that we can squash.

ltratt:
I think it would be best to open a fresh PR.

PR #53, opened 2020-02-13
Add function to specify number of test threads

jacob-hughes:
This function allows the user to specify the default number of test threads ahead of time. This is useful if, for example, a test harness will only work in a single threaded context, since lang_tester attempts to parallelise by default.

It's a draft PR for now as I want to check if you're happy with the API. If so I'll add some docs for it and update it.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -85,6 +85,14 @@ impl<'a> LangTester<'a> {
         self
     }
 
+    /// Specify the number of simultaneous running test cases. Defaults to using
+    /// all available CPUs.
+    pub fn test_threads(&'a mut self, test_threads: usize) -> &'a mut Self {
+        let inner = Arc::get_mut(&mut self.inner).unwrap();
+        inner.test_threads = test_threads;
+        self
+    }
+
     /// If `test_file_filter` is specified, only files for which it returns `true` will be
     /// considered tests. A common use of this is to filter files based on filename extensions
     /// e.g.:


src/tester.rs
```diff
@@ -85,6 +85,18 @@ impl<'a> LangTester<'a> {
         self
     }
 
+    /// Specify the number of simultaneous running test cases. By default, Rust
+    /// will attempt to use all available CPUs.
+    pub fn test_threads(&'a mut self, mut test_threads: usize) -> &'a mut Self {
+        if test_threads > num_cpus::get() {
```
ltratt: I don't think we want this check: if someone wants to specify a large number of test_threads, we should trust them that they have a good reason for doing so (e.g. perhaps many of them wait on the network).

jacob-hughes: That makes sense. I've removed it here: 63a315797f36da4cb4212cbafdf3b3d52253d8f9

src/tester.rs
```diff
@@ -85,6 +85,14 @@ impl<'a> LangTester<'a> {
         self
     }
 
+    /// Specify the number of simultaneous running test cases. By default, Rust
```
ltratt: Maybe change the second sentence to "Defaults to the number of available CPUs."?

jacob-hughes: Fixed: 8c6b2b1280dc51dbfeb7aa0da7f7fa90794d0c41

ltratt:
Please squash.

jacob-hughes:
Are you happy for me to add some docs to the README as part of this PR? I can do this now before squashing if you like.

ltratt:
I'm not sure there's a natural place in the README for this: since it's a Rust API, users will have to read the API documentation anyway.

jacob-hughes:
Okay makes sense. Squashed.

PR #58, opened 2020-05-20
Print out the name of nested tests in a more pleasing way

ltratt:
Although I'd never tried it, lang_tester supported nested tests (i.e. tests in nested directories). However, if you had something like:

```
  lang_tests/
    a.rs
    b/
      c.rs
```

then the output would be:

```
  lang_tests::a.rs
  lang_tests::c.rs
```

That seems OK until you do this:

```
  lang_tests/
    a.rs
    b/
      a.rs
```

and then the output would be:

```
  lang_tests::a.rs
  lang_tests::a.rs
```

Which test is which?

This commit makes it so that nested tests print out their directory name(s):

```
  lang_tests::a.rs
  lang_tests::b::a.rs
```

This does rely on path canonicalisation, which is never entirely guaranteed to succeed, so there is a fallback mechanism if it fails.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -2,13 +2,13 @@ use std::{
     collections::{hash_map::HashMap, HashSet},
     convert::TryFrom,
     env,
-    fs::read_to_string,
+    fs::{canonicalize, read_to_string},
     io::{self, Read, Write},
     os::{
         raw::c_int,
         unix::{io::AsRawFd, process::ExitStatusExt},
     },
-    path::{Path, PathBuf},
+    path::{Path, PathBuf, MAIN_SEPARATOR},
     process::{self, Command, ExitStatus},
     str,
     sync::{
@@ -38,8 +38,7 @@ const INITIAL_WAIT_TIMEOUT: u64 = 10000; // nanoseconds
 /// The maximum time we should wait() between checking if a child process has exited.
 const MAX_WAIT_TIMEOUT: u64 = 250_000_000; // nanoseconds
 
-pub struct LangTester<'a> {
-    test_dir: Option<&'a str>,
+pub struct LangTester {
     use_cmdline_args: bool,
     test_file_filter: Option<Box<dyn Fn(&Path) -> bool>>,
     cmdline_filters: Option<Vec<String>>,
@@ -49,6 +48,7 @@ pub struct LangTester<'a> {
 /// This is the information shared across test threads and which needs to be hidden behind an
 /// `Arc`.
 struct LangTesterPooler {
+    test_dir: Option<PathBuf>,
     test_threads: usize,
     ignored: bool,
     nocapture: bool,
@@ -56,17 +56,17 @@ struct LangTesterPooler {
     test_cmds: Option<Box<dyn Fn(&Path) -> Vec<(&str, Command)> + Send + Sync>>,
 }
 
-impl<'a> LangTester<'a> {
+impl LangTester {
     /// Create a new `LangTester` with default options. Note that, at a minimum, you need to call
     /// [`test_dir`](#method.test_dir), [`test_extract`](#method.test_extract), and
     /// [`test_cmds`](#method.test_cmds).
     pub fn new() -> Self {
         LangTester {
-            test_dir: None,
             test_file_filter: None,
             use_cmdline_args: true,
             cmdline_filters: None,
             inner: Arc::new(LangTesterPooler {
+                test_dir: None,
                 ignored: false,
                 nocapture: false,
                 test_threads: num_cpus::get(),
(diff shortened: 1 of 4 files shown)


src/tester.rs
```diff
@@ -225,22 +231,24 @@ impl<'a> LangTester<'a> {
...
+    /// (`a` and `c`) will be filtered out. The `PathBuf`s returned are guaranteed to be fully
+    /// canonicalised.
     fn test_files(&self) -> (Vec<PathBuf>, usize) {
         let mut num_filtered = 0;
-        let paths = WalkDir::new(self.test_dir.unwrap())
+        let paths = WalkDir::new(self.inner.test_dir.as_ref().unwrap())
             .into_iter()
             .filter_map(|x| x.ok())
             .filter(|x| x.file_type().is_file())
+            .map(|x| canonicalize(x.into_path()).unwrap())
```
jacob-hughes, line 242: Do we `unwrap` here because the paths are expected to have already been passed to `canonicalize` before (and hence, this is guaranteed not to fail)?

ltratt: This isn't guaranteed to fail, but I've never seen one of these calls fail, so I'm not really sure what to do / report when it does fail. In retrospect, `lang_tester` should probably use `Result` in return types more often, but I'm not sure it's worth changing the API now...

src/tester.rs
```diff
@@ -475,7 +482,19 @@ fn test_file(
     let num_ignored = Arc::new(AtomicUsize::new(0));
     let pool = ThreadPool::new(inner.test_threads);
     for p in test_files {
-        let test_fname = p.file_stem().unwrap().to_str().unwrap().to_owned();
+        let test_fname = loop {
```
jacob-hughes, line 485: `test_file()` is getting pretty long now. Do you think it might be worth breaking this into a helper function?

src/tester.rs
ltratt: "isn't guaranteed not to fail" even...

jacob-hughes: Guarantee was a poor word choice. What I meant was on line 84 you have:
```rust          
inner.test_dir = Some(canonicalize(test_dir).expect(&format!(...
```

If this is always run first, then you're expecting `canonicalize` not to fail the second time around too?

If that's not the case, perhaps this `unwrap` should be replaced by `expect` in the same way.

ltratt: Agreed, it *probably* won't fail. In general, I'm not sure `expect` is better than `unwrap`: it makes the code uglier, often for cases that we'll probably never see. I suspect this is one of them.

jacob-hughes:
Looks good to me, a couple of minor comments. Will you want to squash or should those commits remain separate?

src/tester.rs
jacob-hughes: Agreed. On that basis do you think that the `expect` on line 84 should just be an `unwrap` too?

ltratt: Good question. I assumed that there's a small possibility that one might fail, but I'm just as happy for it to be an `unwrap`! Thoughts?

jacob-hughes: I prefer `unwrap` tbh. I'm not sure the `expect` adds much apart from ugliness - In the rare case it does fail, `unwrap` will give a line number and it will be easy to then see that it went wrong in some way.

jacob-hughes:
Actually one other thing. The commit message mentions a fallback mechanism if canonicalisation fails, but I can't see one? If we `unwrap` an error surely that test thread will just panic?

src/tester.rs
ltratt: Fixed in https://github.com/softdevteam/lang_tester/pull/58/commits/a658fd03487f922ff73d14c256148f72ed25e3c4.

src/tester.rs
ltratt: Fixed in https://github.com/softdevteam/lang_tester/pull/58/commits/340709b36e1a0f85aa210934ffa14ee50feb5751.

ltratt:
> The commit message mentions a fallback mechanism if canonicalisation fails, but I can't see one?

Ah, the message is slightly misleading. If you look at this part of the `test_fname` function:

```rust
    if let Some(test_fpath) = test_fpath.as_os_str().to_str() {
        if let Some(testdir_path) = test_dir_path.as_os_str().to_str() {
            if test_fpath.starts_with(testdir_path) {
                return test_fpath[testdir_path.len() + MAIN_SEPARATOR.len_utf8()..]
```

you'll see that we rely on canonicalisation giving totally predictable strings where we can identify common prefixes, which isn't exactly guaranteed. So it's not the canonicalisation call fails, but that we get something back that doesn't quite fit our ideal scheme. I'll update the message accordingly if/when we get to squashing.

jacob-hughes:
Aha I see. Ok, LGTM. Please squash.

ltratt:
Squashed.

PR #62, opened 2020-07-03
Remove lang_tester's fuzzy matcher and use fm instead.

ltratt:
This PR minimally moves lang_tester to use https://github.com/softdevteam/fm/. This should give exactly the same externally observable behaviour, but it also gives us the opportunity to make use of fm's extra features.


vext01:
Well, that was straightforward!

Since CI is down, I've done rustfmt and testing locally. All seems well.

Shall I manually merge?

ltratt:
If you're OK with merging, please do!

PR #63, opened 2020-07-04
Add the ability to override fm's defaults when using lang_tester.

ltratt:
This allows lang_tester users to e.g. use name matching in their tests.

I'm not sure if this is the right API or not: it's flexible but that brings some complexity as well. We could, for example, pass a boolean `is_stderr`, but that would restrict us forever more to just stderr/stdout tests. Maybe that's all we need, but I'm not sure.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -19,7 +19,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use fm::FMatcher;
+use fm::FMBuilder;
 use getopts::Options;
 use libc::{fcntl, poll, pollfd, F_GETFL, F_SETFL, O_NONBLOCK, POLLERR, POLLHUP, POLLIN};
 use termcolor::{Color, ColorChoice, ColorSpec, StandardStream, WriteColor};
@@ -54,7 +54,17 @@ struct LangTesterPooler {
     ignored: bool,
     nocapture: bool,
     test_extract: Option<Box<dyn Fn(&str) -> Option<String> + Send + Sync>>,
+    fm_options: Option<
+        Box<dyn for<'a> Fn(&'a Path, TestStream, FMBuilder<'a>) -> FMBuilder<'a> + Send + Sync>,
+    >,
     test_cmds: Option<Box<dyn Fn(&Path) -> Vec<(&str, Command)> + Send + Sync>>,
+}
+
+/// Specify a given test stream.
+#[derive(Clone, Copy)]
+pub enum TestStream {
+    Stderr,
+    Stdout,
 }
 
 impl LangTester {
@@ -71,6 +81,7 @@ impl LangTester {
                 ignored: false,
                 nocapture: false,
                 test_threads: num_cpus::get(),
+                fm_options: None,
                 test_extract: None,
                 test_cmds: None,
             }),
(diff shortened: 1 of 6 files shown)


ltratt:
bors try

ltratt:
bors try-

ltratt:
bors try

ltratt:
@jacob-hughes I'd also be interested in your comments here!

src/tester.rs
```diff
@@ -146,6 +157,31 @@ impl LangTester {
...
+    ///    ...
+    ///    .fm_options(|_, _, fmb| {
+    ///        let ptn_re = Regex::new(r"\$.+?\b").unwrap();
+    ///        let text_re = Regex::new(r".+?\b").unwrap();
+    ///        fmb.name_matcher(Some((ptn_re, text_re)))
+    ///    })
+    /// ```
+    pub fn fm_options<F>(&mut self, fm_options: F) -> &mut Self
+    where
+        F: 'static + for<'a> Fn(&'a Path, TestStream, FMBuilder<'a>) -> FMBuilder<'a> + Send + Sync,
```
vext01, line 179: Do we need to document the first two parameters of the closure? There's a path and a `TestStream` that I'm unsure what are for.

ltratt: You're right, they should be documented, but if you can't guess what they are with 100% accuracy (once you've looked up what `TestStream` is), I'd be surprised ;)

ltratt: Actually they are documented!

```
This function can be used to override `fm`'s defaults for a given file, a given testing stream (stderr or stdout)
```

vext01:
This looks pretty reasonable to me, you just forward the `FmBuilder` up to lang_tester.

The only downside I can see is that if you do an ABI bump on `fm` then `lang_tester` also needs one too. It could be easy to forget?

I didn't understand the comment about `is_stderr`.

vext01:
> The only downside I can see is that if you do an API bump on fm then lang_tester also needs one too. It could be easy to forget?

Actually, as long as we do semver correctly, I think we might be OK?

ltratt:
> The only downside I can see is that if you do an ABI bump on fm then lang_tester also needs one too. It could be easy to forget?

No, lang_tester can just stick with the older versions of `fm` if it wants (assuming that `fm` respects semver!).

> I didn't understand the comment about is_stderr.

Instead of `TestStream` (which currently is *only* stderr/stdout) we could just pass a boolean. But that assumes we'll only ever have stderr/stdout.

vext01:
> Instead of TestStream (which currently is only stderr/stdout) we could just pass a boolean. But that assumes we'll only ever have stderr/stdout.

Seems reasonable to me.

ltratt:
Reasonable as-is or changed to a boolean?

vext01:
As-is. If I understand correctly, this is more flexible and would allow us to do matches on (say) files on the filesystem.

ltratt:
OK, then we're on the same page.

Are we OK to merge? I'm not sure Jake will see this PR any time soon TBH.

src/tester.rs
vext01: > This function can be used to override `fm`'s defaults for a given file, a given testing stream

So that's a file *and* a testing stream (since they are not `Option`). So what path would you pass for e.g. stdout? Ideally it should be clear from this doc-string without having to look elsewhere.

ltratt: Does https://github.com/softdevteam/lang_tester/pull/63/commits/f43fb87b841ede24dce246152f1ce28e6566faaa help?

vext01: Ah, so the `Path` is to an executable which is run to get stderr/stdout?

I see.

ltratt: No, it's the test file (exactly the same as for `test_cmds` and so on).

vext01: I think I get it, but was confused by:
> when executing that file

I guess someone/something is compiling the test file. As far as I can see they are just Rust programs with special comments, which can't be directly executed. Maybe I shouldn't take it so literally.

ltratt: Try https://github.com/softdevteam/lang_tester/pull/63/commits/0e37c052eb45cfd434fed34c2da135786a2fc931.

vext01: OK.

Sorry if i was pedantic. I've never used lang_tester, so I'm probably out of the loop.

vext01:
Please squash.

ltratt:
Squashed.

src/tester.rs
ltratt: Not a problem!

PR #64, opened 2020-07-06
Add the ability to specify stdin to pass to a sub-command.

ltratt:


src/lib.rs
```diff
@@ -100,6 +100,10 @@
 //!
 //!   * `extra-args: <arg 1> [... <arg n>]`, where each space separated argument will be appended,
 //!     in order, to those arguments specified as part of the `test_cmds` function.
+//!   * `stdin: <string>`, text to be passed to the command's `stdin`. If the command exits without
+//!     having consumed all of `<string>`, an error will be raised. Note, though, that operating
+//!     system file buffers can mean that the command *appears* to have consumed all of `<string>`
+//!     without it actually having done so.
```
jacob-hughes, line 106: E.g. You have a bunch of newlines in your input string?

ltratt:
Oops, forgot to add @vext01 to this.

src/lib.rs
ltratt: The problem is even simpler than that: the OS and/or filesystem library might buffer input. So we might write to the pipe, see that the input has disappeared, and assume that it was read by the command; but the input might be in an intermediate buffer and the program can exit without ever actually having read it. I'm not aware that we can do anything about this.

vext01:
LGTM

ltratt:
OK, so it sounds like we can merge?

vext01:
There's one outstanding comment from Jake, so I'll let him do the merge once he's happy.

ltratt:
@vext01 I think we might want to, and be able to, go ahead with this one. Please have a look at my reply to Jake's question and see if you think it's been addressed or not.

src/lib.rs
vext01: That matches my interpretation of the docstring.

jacob-hughes: Ah I understand. Thanks!

Issue #65, opened 2020-07-07
Multi-line stdin


ltratt:
What should the following do?

```
VM:
  stdin: x
```

It seems fairly clear to me that it should pass the string `x` to the command. Notice that we've stripped the space character between `:` and `x`.

What should we do with multi-line stdin definitions such as:

```
VM:
  stdin:
    x
    y
```

At the moment we would pass the string `x\ny` to the command i.e. we strip the whitespace before `y` too. I believe that's the least-surprising thing to do, but we should consider this before we make a release and start to rely on one behaviour over another.

Thoughts @vext01?

vext01:
I may not be the best person to ask, having never used lang tester, but here goes.

> Notice that we've stripped the space character between : and x

I see a space still.

> At the moment we would pass the string x\ny

So is this about whether or not to strip the indentation?

ltratt:
> So is this about whether or not to strip the indentation?

In essence, yes.

vext01:
We could do [as Rust does](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=152406667bc97d2bace4e373c26bcb64)? i.e. you can have either behaviour depending on if you use `\` or not.

ltratt:
Interesting idea. I might need to chew on this for a bit.

PR #66, opened 2020-07-08
Don't strip leading whitespace from lines.

ltratt:
This allows us to write things such as:

```
  x:
    a
      b
    c
```

and differentiate the leading whitespace in the lines. Trailing whitespace is always stripped. Note that this change is backwards compatible because `fm` defaults to whitespace-insensitive matching, so unless users explicitly request whitespace-sensitive matching they won't know the difference.

This fixes https://github.com/softdevteam/lang_tester/issues/65 in a simpler way than I think we originally thought possible.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ name = "fm_options"
 path = "examples/fm_options/run_tests.rs"
 
 [dependencies]
-fm = "0.1"
+fm = "0.1.1"
 getopts = "0.2"
 libc = "0.2"
 num_cpus = "1.10"
--- a/README.md
+++ b/README.md
@@ -81,8 +81,21 @@ fn main() {
 Test data is specified with a two-level indentation syntax: the outer most
 level of indentation defines a test command (multiple command names can be
 specified, as in the above); the inner most level of indentation defines
-alterations to the general command or sub-tests. Each test command must
-define at least one sub-test:
+alterations to the general command or sub-tests. Multi-line values are stripped
+of their common indentation, such that:
+
+```text
+x:
+  a
+    b
+  c
+```
+
+defines a test command `x` with a value `a\n  b\nc`. Trailing whitespace is
+not stripped from each line. Note that by default `fm` ignores leading and
+trailing whitespace.
+
+Each test command must define at least one sub-test:
 
   * `status: <success|failure|signal|<int>>`, where `success` and `failure` map
     to platform specific notions of a command completing successfully or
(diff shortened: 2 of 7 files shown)


ltratt:
Now that I think about it: should we strip the trailing whitespace? Though if we don't, does that mean "keep the newline" characters? It is a bit tricky to reason about, so perhaps stripping the trailing whitespace is the easiest things to do...

vext01:
I'm not sure I follow.

Suppose the language gave meaning to leading and trailing whitespace. Would I still be able to test such a language?

ltratt:
> Suppose the language gave meaning to leading and trailing whitespace. Would I still be able to test such a language?

I think that's not quite the right way to think about this. Rather: you tell lang_tester how to extract a test (probably from a comment) but assuming you give it leading/trailing whitespace, it can then preserve them if it wants to. Whether the language is whitespace sensitive or not is a red herring. Notice that we have a Python test in this PR!

vext01:
> Multi-line values are stripped of their common indentation

Maybe I'm confused, but this would mean that you *couldn't* reliable test the indent of (e.g.) python source code.

Suppose I'm looking for (exactly):
```
    while(true):
        pass
```

And I really care that the first line is indented by four chars, I don't think I'd be able to express it, would I?

ltratt:
Have you looked at the Python example in the PR?

vext01:
Yes, but that doesn't answer the question, as the first line has no leading whitespace.

vext01:
Put differently, if I want to match a line `<space><space><space><space>while True:` and I want it to fail is there are not four spaces, is that possible?

ltratt:
The `echo.py` example is per-line whitespace sensitive (notice the change in `fm_options/run_tests.rs`). You can try it out locally if you want: removing the whitespace on the `b` line in either the pattern or text will cause a non-match.

vext01:
> Now that I think about it: should we strip the trailing whitespace?

I think that'd be a sensible default, as long as it's documented. The user can always tweak the FM options, if they don't like it, right?

ltratt:
https://github.com/softdevteam/lang_tester/pull/66/commits/e897e57afee187ebf6aa3adeeddc172a3656073d also strips trailing whitespace. I think this makes things more symmetric, if someone has the (not very likely, but not totally impossible) situation where they want matching of trailing whitespace. `fm` ignores this by default so it has no observable behaviour unless people change `fm`'s options.

vext01:
OK, I think we can move on with this.

Any squashing to do?

ltratt:
Squashed.

PR #68, opened 2020-07-13
Display errors as fm output.

ltratt:
If an fm match fails, this commit uses the output from fm rather than printing
the raw stdout/stderr. For example, with the upcoming context support in fm,
this leads to the helpful output such as the following:

```
  ---- lang_tests::nested_backtrace2.som stderr ----

  Pattern (error at line 3):
     |Traceback (most recent call at bottom):
     |  ...
  >> |  ...nested_backtrace2.som, line 30, column 12:
     |      run = ( self m )
     |  ...nested_backtrace2.som, line 26, column 8:
     |          1 == 0
     ...

  Text (error at line 28):
     ...
     |          self ifFalse: [ ^falseBlock value ].
     |  File /home/ltratt/yksom/lang_tests/nested_backtrace2.som, line 29, column 21:
     |            ifFalse: [ 2 / 0 ].
  >> |Division by zero.
```

This makes it much easier to understand why a test failed. Previously I had
sometimes spent ridiculous periods of time trying to work out why a test pattern
had not matched: I think this will cut that down significantly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ name = "fm_options"
 path = "examples/fm_options/run_tests.rs"
 
 [dependencies]
-fm = "0.1.1"
+fm = "0.1.2"
 getopts = "0.2"
 libc = "0.2"
 num_cpus = "1.10"
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -19,7 +19,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use fm::FMBuilder;
+use fm::{FMBuilder, FMatchError};
 use getopts::Options;
 use libc::{
     close, fcntl, poll, pollfd, F_GETFL, F_SETFL, O_NONBLOCK, POLLERR, POLLHUP, POLLIN, POLLOUT,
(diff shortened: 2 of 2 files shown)


vext01:
Can't argue with that.

bors r+

ltratt:
Oh, sausages, we need to make an fm release first.

vext01:
I've added buildbot as a required check, so next time we will get a link to the build log.

For the last failure, it's here:
http://ci.soft-dev.org:8010/#/builders/1/builds/585/steps/2/logs/stdio

We need to use a version of fm that has the error equality?

ltratt:
> We need to use a version of fm that has the error equality?

Yes. I forgot that bit! Once that release is made, https://github.com/softdevteam/lang_tester/pull/68/commits/5a0a1268792cfba1ae45598c7c91b87f894319af will ensure that you can't use this with a version of fm without that suppotr.

vext01:
> Once that release is made, 5a0a126 will ensure that you can't use this with a version of fm without that support.

Checked the semver rules to be sure. Yes, I agree.

vext01:
sorry, please squash.

ltratt:
Squashed and fm updated.

bors try

PR #70, opened 2020-10-31
Clearly document that fm is used for string matching.

ltratt:


README.md
```diff
@@ -91,8 +91,13 @@ x:
   c
 ```
 
-defines a test command `x` with a value `a\n  b\nc`. Trailing whitespace is
-not stripped from each line. Note that by default `fm` ignores leading and
+defines a test command `x` with a value `a\n  b\nc`. Leading whitespace is thus
+implicitly removed; however, trailing whitespace is preserved.
+
+String matching is performed by the [fm crate](https://crates.io/crates/fm),
+which provides support for `...` operators and so on. Unless `lang_tester` is
```
vext01, line 98: Something is weird here.

> Unless `lang_tester` is explicitly instructed otherwise, it uses `fm`'s defaults.

OK

> Thus, for example, while `lang_tester` preserves trailing whitespace, `fm`'s default is to ignore trailing whitespace.

But you just said that by default it does the same as fm unless the user says otherwise? Here you are saying it does something different by default.

Do I misunderstand?

ltratt: Yes, you do misunderstand, but it's understandable that you misunderstand! `lang_tester` uses `fm` for text matching, so you have this slightly odd situation where `lang_tester` preserves trailing whitespace but `fm` ignores it. Explaining this clearly is rather hard, and I'm open to suggestions as to possible improvements (but I'm also a little fatalistic that the above text might be the best of a bad job).

vext01: Does wording like: "Aside from stripping trailing whitespace by default, lang_tester uses fm defaults unless explicitly overridden"

Or similar.

ltratt: > Aside from stripping trailing whitespace by default, lang_tester uses fm defaults unless explicitly overridden

Unfortunately that's not what happens. lang_tester *never* strips trailing whitespace. *But* `fm` ignores trailing whitespace by default.

vext01: I'm so confused. Can you try revising the text, because as it stands the prose seems to conflict with itself.

* lang tester uses fm defaults
* but lang tester doesn't use fm defaults.

ltratt: > lang tester uses fm defaults

This is correct. I can't improve on my explanation above, so I'm going to be rude, and ask if you can try rereading it with a fresh mind. These are the facts:

  * lang_tester uses's fm defautls.
  * lang_tester doesn't strip trailing whitespace.
  * fm ignores trailing whitespace when matching.

ltratt: Ping @vext01.

vext01: > fm ignores trailing whitespace when matching.

So now for a dumb question. Is that the same as saying that fm *does* strip trailing whitespace?

ltratt: Ignoring/stripping are the same thing in this case IMHO.

vext01: OK. So let's substitute that bullet point. We have:

 * lang_tester uses fm's defaults.
 * fm does strip trailing whitespace [by default]
 * lang_tester doesn't strip trailing whitespace.

Do you see the confusion? These statements are inconsistent. If lang tester doesn't strip whitespace it must have deviated from fm defaults.

I won't die on this hill, but I don't get it :P

ltratt: > If lang tester doesn't strip whitespace it must have deviated from fm defaults.

Ah, I think maybe the relationship between lang_tester and fm isn't clear.

lang_tester isn't doing string matching, but it is tearing strings apart: it hands the torn-apart strings to fm for matching.

Does that make sense? There's nothing inconsistent here, though I admit that the relationship is pretty subtle :/

vext01: Alright, if it's at different levels in the system, then I'm probably talking rubbish.

Let's move on.

vext01:
So this is ready to go?

README.md
ltratt: See if https://github.com/softdevteam/lang_tester/pull/70/commits/28a6b6709266be789f1b06d9a6e4cf9c6068a7b2 makes a bit more sense.

vext01: Honestly it's about the same for me.

I don't think we should obsess on this.

ltratt: I think I prefer 28a6b67, so let's maybe squash it down?

vext01:
Please squash.

ltratt:
Squashed.

Issue #76, opened 2021-01-25
Support out-of-file stdout / stderr output.


xldenis:
The current API for `test_extract` requires the `stdout/stderr` output to be stored in the test file. This is rather inconvenient if we wish to update these outputs programatically. Instead what is commonly done (for example in `compiletest`) is to store the outputs in `*.stdout` and `*.stderr` files alongside the test cases. 

Would you be open to a secondary (or replacement) API which enables this convention? 

One possible API could be:

```
fn test_extract_path(extractor: Option<Box<dyn Fn(&Path) -> Tests + Send + Sync>>)
```

of course `test_extract` could be expressed as a wrapper around `test_extract_path`. 

Test case regeneration could either be provided by `lang_tester` directly when `test_extract_path` is set, or by test authors overwriting files manually in their tests. If `lang_tester` provides this then `Tests` should store the expectation `Path`s rather than the file contents.

ltratt:
I'm potentially open to this but we'd need to find a non-annoying way of doing so. Two problems jump out at me:

1. Ideally, we don't want to allow people to express both `test_extract` and `test_extract_path`. That suggests collapsing them into one function but ...
2. We don't really want to muck around with requiring say an `enum` for different test options.

I'm not immediately sure how to resolve this quandary but I'll have a think.

xldenis:
> I'm potentially open to this but we'd need to find a non-annoying way of doing so. Two problems jump out at me:
> 
>     1. Ideally, we don't want to allow people to express both `test_extract` and `test_extract_path`. That suggests collapsing them into one function but ...

I think that `test_extract` could be a wrapper around `test_extract_path` that performs a file read and passes that to the closure. The regeneration of output files could be a flag that is set as off in `test_extract`. 
> 
>     2. We don't really want to muck around with requiring say an `enum` for different test options.

Why not? It's certainly better than passing a string in.
> 
> 
> I'm not immediately sure how to resolve this quandary but I'll have a think.

ltratt:
Having played around with this, I think `test_extract` should simply take a `&Path`. Please have a look at https://github.com/softdevteam/lang_tester/pull/77 and see what you think -- comments are welcome!

PR #77, opened 2021-01-25
Make test_extract take `&Path`s rather than `&str`s.

ltratt:
lang_tester tries to make it as easy as possible to set-up, so it has a fairly simple API. One way I did that was by passing the contents of a file to a `test_extract` function. However, that is fairly inflexible, and also it turns out not very much simpler (it doesn't help that `test_extract` returns `Option<String>`: frankly, it's easier for tests to panic in such situations).

This commit therefore changes the `test_extract` function signature from:

  ```
  Fn(&str) -> Option<String> + Send + Sync,
  ```

  to:

  ```
  Fn(&Path) -> String + Send + Sync,
  ```

In other words, users now have to a) read the contents of a path themselves (but
it doesn't necessarily have to be the path passed to the function!) b) return a
`String` rather than an `Option<String>`.

In practise, most `test_extract` functions can be changed from (roughly):

  ```
  test_extract(|s| { s.lines() ... })
  ```

to:

  ```
  test_extract(|p| { read_to_string(p).lines() })
  ```

Fixes https://github.com/softdevteam/lang_tester/issues/76 though at the expense of a relatively painful API bump. I think it's worth it though.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,31 @@
+# lang_tester 0.5.0 (xxxx-xx-xx)
+
+* The `test_extract` function signature has changed from:
+  ```
+  Fn(&str) -> Option<String> + Send + Sync,
+  ```
+  to:
+  ```
+  Fn(&Path) -> String + Send + Sync,
+  ```
+
+  In other words, users now have to both:
+
+    1. Read the contents of a path themselves (but it doesn't necessarily have
+       to be the path passed to the function!)
+    2. Return a `String` rather than an `Option<String>`.
+
+  In practise, most `test_extract` functions can be changed from (roughly):
+  ```
+  test_extract(|s| { s.lines() ... })
+  ```
+  to:
+  ```
+  test_extract(|p| { read_to_string(p).lines() })
+  ```
+
+
+
 # lang_tester 0.4.0 (2020-11-26)
 
 * Update to fm 0.2.0. This changes the interface exposed by the `fm_options`
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ For example, a Rust language tester, loosely in the spirit of
 [`compiletest_rs`](https://crates.io/crates/compiletest_rs), looks as follows:
 
 ```rust
-use std::{path::PathBuf, process::Command};
+use std::{fs::read_to_string, path::PathBuf, process::Command};
 
 use lang_tester::LangTester;
 use tempfile::TempDir;
@@ -26,17 +26,17 @@ fn main() {
         // Only use files named `*.rs` as test files.
         .test_file_filter(|p| p.extension().unwrap().to_str().unwrap() == "rs")
         // Extract the first sequence of commented line(s) as the tests.
-        .test_extract(|s| {
-            Some(
-                s.lines()
+        .test_extract(|p| {
+            read_to_string(p)
+                .unwrap()
+                .lines()
                     // Skip non-commented lines at the start of the file.
                     .skip_while(|l| !l.starts_with("//"))
                     // Extract consecutive commented lines.
                     .take_while(|l| l.starts_with("//"))
                     .map(|l| &l[2..])
                     .collect::<Vec<_>>()
-                    .join("\n"),
-            )
+                    .join("\n")
         })
         // We have two test commands:
         //   * `Compiler`: runs rustc.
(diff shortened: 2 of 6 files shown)


CHANGES.md
```diff
@@ -1,3 +1,29 @@
...
+  Fn(&str) -> Option<String> + Send + Sync,
+  ```
+  to:
+  ```
+  Fn(&Path) -> String + Send + Sync,
+  ```
+
+  In other words, users now have to a) read the contents of a path themselves
+  (but it doesn't necessarily have to be the path passed to the function!) b)
+  return a `String` rather than an `Option<String>`.
```
jacob-hughes, line 14: Can we try rephrasing this? It took me a couple of re-reads to parse correctly as I read the "a)" and "b)" as mutually exclusive options. Even just switching to _"users now have to do two things: ...."_ would help.

jacob-hughes:
Minor comment about phrasing in the README, otherwise LGTM. I think I prefer this API.

CHANGES.md
ltratt: Is https://github.com/softdevteam/lang_tester/pull/77/commits/937290345004d94d4b080aa0e6f66002b397ac31 enough or is more needed?

jacob-hughes: Yes that's much better. Can you change it in the commit message too when you squash.

jacob-hughes:
Pleas squash

CHANGES.md
ltratt: Will do!

ltratt:
I'd like to wait a day or so to see if @xldenis feels this solves the issue he raised. If it does, I'll squash (and fix the commit message at the same time).

jacob-hughes:
Sure

xldenis:
This doesn't solve the underlying issue of #76 in my opinion as it still has a `String` output type. 

The use case I'm envisioning is to have `.stdout` and `.stderr` alongside my tests and use those as the expected output. 

It seems like if I want to use this in the current api I need to construct a string that corresponds to the test DSL header out of those files? 

I see two real solutions: structure the output type, or expose a lower level primitive which asks users to directly construct a `Tests` struct.

ltratt:
I think there's a fairly simple work around. Your `test_extract` function can read `x.stderr` and `x.stdout` and combine them into a `String` which looks as follows:

```
stdout:
  ...
stderr:
  ...
```

Does that sound doable?

xldenis:
@ltratt it is doable but it feels like a less satisfactory api if I'm honest, why pass around unstructured types when we can have structure?

xldenis:
What is the problem with using something like:

```
pub fn test_extract_prim<F>(&mut self, test_extract: F) -> &mut Self
    where
        F: 'static + Fn(&Path) -> Tests + Send + Sync,
```
?

The existing `test_extract` would just take a user closure returning a `String` wrap it, calling `parse_tests` and pass it along.

ltratt:
@xldenis Structure comes at an ease-of-use cost: your API suggestion is nice in one way but now I need to expose `Tests` to the user and they have to understand it, and it increases the complexity. It's always a trade-off and lang_tester has been designed from the start to slightly favour ease-of-use over flexibility. However, I think your use case is something that lang_tester should be able support, hence this PR and the API change it necessitates. Your use case might not be supported in quite the way that you want, but at least we can offer you *something* now, which I hope is better than nothing :)

@jacob-hughes Squashed.

xldenis:
Of course, I'll take this improvement! 

 I'm not convinced that an adhoc DSL is simpler than a structured type though, users still need to understand the structure and limitations of your parser.

ltratt:
@xldenis There are definitely situations where lang_tester isn't sufficient: from the very beginning, I wanted to get 80% of the power of https://crates.io/crates/compiletest_rs with 20% of the complexity, and I think we're in that sort of range. There's probably a place for a much more flexible, but complex, language neutral equivalent of lang_tester / compiletest_rs, but we've been pleasantly surprised at how many projects we've been able to slot lang_tester into without issue.

xldenis:
Fair enough! One final option before I let you go! Define a few `From` instances for `Tests` and use those as the output type.
You could have a `From<String>` which uses `parse_tests` and a `From<HashMap<String, (String, String)>` or whatever.

ltratt:
Interesting idea. Let me chew on it!

jacob-hughes:
@ltratt I don't think the commit message was updated?

ltratt:
Mea culpa! Force pushed an update.

PR #79, opened 2021-04-29
Try to get some of the concepts across more clearly in the docs

ltratt:
This PR tries to clarify the documentation somewhat. The problem with `lang_tester` is that it's so brutally simple that there's a tendency to assume that it can't possibly be *that* simple. I've tried to bring out a few of the concepts more explicitly to emphasise the fundamentals a bit better. This will probably never be perfect, but anything which makes it better can only be a good thing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ fn main() {
                     .skip_while(|l| !l.starts_with("//"))
                     // Extract consecutive commented lines.
                     .take_while(|l| l.starts_with("//"))
+                    // Strip the initial "//" from commented lines.
                     .map(|l| &l[2..])
                     .collect::<Vec<_>>()
                     .join("\n")
--- a/README.md
+++ b/README.md
@@ -17,6 +17,8 @@ use std::{fs::read_to_string, path::PathBuf, process::Command};
 use lang_tester::LangTester;
 use tempfile::TempDir;
 
+static COMMENT_PREFIX: &str = "//";
+
 fn main() {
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -34,6 +34,7 @@ fn main() {
                     .skip_while(|l| !l.starts_with("//"))
                     // Extract consecutive commented lines.
                     .take_while(|l| l.starts_with("//"))
+                    // Strip the initial "//" from commented lines.
```
vext01, line 37: Why don't we define `//` as a `const LANG_TEST_COMMENT`, replace the duplicated strings above, and then do `.map(|l| &l[LANG_TEST_COMMENT.len()..])`?

This way the comment can be changed in one place.

ltratt: I'm not sure that makes things clearer TBH.

vext01:
Just that one comment.

README.md
vext01: My plunder earlier is evidence that it would be better :)

To change the comment here, I'd have to update 3 places, one of which can't be detected by searching for the comment string.

ltratt: "Plunder" is a wonderfully Freudian typo ;)

I take your point though. Fixed in https://github.com/softdevteam/lang_tester/pull/79/commits/109f597f923a95c53e09501838f1d35533f366a3.

vext01:
LGTM. Squash if ready.

ltratt:
Squashed.

PR #80, opened 2021-04-29
Report errors in the test runner as if they were test runners.

ltratt:
Before this commit, if the user made a panic-inducing mistake in, for example, `test_extract` then they will see text to stderr, but the test counts as successful.

This commit uses the catch_unwind interface to detect such mistakes and count them as test failures. This can sometimes be a little counter-intuitive (for example, panicing inside `test_files` now counts as a test failure, even though the test hasn't been run), but it does mean that these can't scroll by harmlessly. Using catch_unwind forces us to make the functions passed to lang_tester be RefUnwindSafe, which will require a major API change (though, given the nature of the sorts of functions people are likely to use with lang_tester, few are likely not to be unwind safe).

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -1,4 +1,5 @@
 use std::{
+    cmp::max,
     collections::{hash_map::HashMap, HashSet},
     convert::TryFrom,
     env,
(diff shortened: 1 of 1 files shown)


vext01:
Looks good, but can we add a test?

ltratt:
> can we add a test?

Not easily in the current setup.

vext01:
OK. bors r+

ltratt:
The `bors` bit needs to be on a new line.

vext01:
bors try

vext01:
OK to splat?

vext01:
Sorry, it appears I am confused :P

bors r+

PR #83, opened 2021-04-30
Allow multiple test directories to be specified.

ltratt:
@vext01 is this what you wanted? Please feel free to test that the functionality actually works as I've only tried it with one directory!

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -53,7 +53,7 @@ pub struct LangTester {
 /// This is the information shared across test threads and which needs to be hidden behind an
 /// `Arc`.
 struct LangTesterPooler {
-    test_dir: Option<PathBuf>,
+    test_dirs: Vec<PathBuf>,
     test_threads: usize,
     ignored: bool,
     nocapture: bool,
@@ -86,7 +86,7 @@ impl LangTester {
             use_cmdline_args: true,
             cmdline_filters: None,
             inner: Arc::new(LangTesterPooler {
-                test_dir: None,
+                test_dirs: Vec::new(),
                 ignored: false,
                 nocapture: false,
                 test_threads: num_cpus::get(),
@@ -97,12 +97,13 @@ impl LangTester {
         }
     }
 
-    /// Specify the directory where test files are contained. Note that this directory will be
+    /// Specify a directory where test files are contained. Note that this directory will be
     /// searched recursively (i.e. subdirectories and their contents will also be considered as
-    /// potential test files).
+    /// potential test files). Calling this function multiple times adds further directories to the
+    /// list of those searched.
     pub fn test_dir(&mut self, test_dir: &str) -> &mut Self {
         let inner = Arc::get_mut(&mut self.inner).unwrap();
-        inner.test_dir = Some(canonicalize(test_dir).unwrap());
+        inner.test_dirs.push(canonicalize(test_dir).unwrap());
         self
     }
 
@@ -267,8 +268,8 @@ impl LangTester {
 
     /// Make sure the user has specified the minimum set of things we need from them.
     fn validate(&self) {
-        if self.inner.test_dir.is_none() {
-            fatal("test_dir must be specified.");
+        if self.inner.test_dirs.is_empty() {
+            fatal("One or more test directories must be specified.");
         }
         if self.inner.test_extract.is_none() {
             fatal("test_extract must be specified.");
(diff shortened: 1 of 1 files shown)


vext01:
Would you like to include this in the release in the other PR? Which PR should I look at first?

ltratt:
Let's do this one, as then we can do 1 release with both features.

vext01:
To test this I copied `examples/rust_lang_tester/lang_tests` to `examples/rust_lang_tester/more_lang_tests` and applied this diff:

```diff
diff --git a/examples/rust_lang_tester/run_tests.rs b/examples/rust_lang_tester/run_tests.rs
index 4d56810..4f502fb 100644
--- a/examples/rust_lang_tester/run_tests.rs
+++ b/examples/rust_lang_tester/run_tests.rs
@@ -17,6 +17,7 @@ fn main() {
     let tempdir = TempDir::new().unwrap();
     LangTester::new()
         .test_dir("examples/rust_lang_tester/lang_tests")
+        .test_dir("examples/rust_lang_tester/more_lang_tests")
         // Only use files named `*.rs` as test files.
         .test_file_filter(|p| p.extension().unwrap().to_str().unwrap() == "rs")
         // Extract the first sequence of commented line(s) as the tests.
```

This makes non-deterministic failures:
```
vext01@bencher13:~/research/lang_tester$ cargo run --example rust_lang_tester
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/examples/rust_lang_tester`

running 20 tests
test lang_tests::ignore.rs ... ignored
test lang_tests::ignore.rs ... ignored
test lang_tests::no_main.rs ... ok
test lang_tests::unknown_var.rs ... ok
test lang_tests::unused_var.rs ... FAILED
test lang_tests::no_main.rs ... ok
test lang_tests::unknown_var.rs ... ok
test lang_tests::custom_cla.rs ... FAILED
Fatal exception:
  Couldn't run command "/tmp/.tmpfz216Z/sig_caught".
vext01@bencher13:~/research/lang_tester$ cargo run --example rust_lang_tester
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/examples/rust_lang_tester`

running 20 tests
test lang_tests::ignore.rs ... ignored
test lang_tests::ignore.rs ... ignored
test lang_tests::no_main.rs ... ok
test lang_tests::no_main.rs ... ok
test lang_tests::unknown_var.rs ... ok
test lang_tests::unknown_var.rs ... ok
test lang_tests::echo.rs ... FAILED
test lang_tests::echo_multiline.rs ... FAILED
Fatal exception:
  Couldn't run command "/tmp/.tmphhPT9N/exit_code".
vext01@bencher13:~/research/lang_tester$ cargo run --example rust_lang_tester
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/examples/rust_lang_tester`

running 20 tests
test lang_tests::ignore.rs ... ignored
test lang_tests::ignore.rs ... ignored
test lang_tests::no_main.rs ... ok
test lang_tests::no_main.rs ... ok
test lang_tests::unknown_var.rs ... ok
test lang_tests::unknown_var.rs ... ok
test lang_tests::echo.rs ... FAILED
Fatal exception:
  Couldn't run command "/tmp/.tmp9AxFVz/sig_caught".
```

Could it be due to the filenames in the two dirs clashing?

ltratt:
Hmm, dunno. I'll look into it.

ltratt:
Are you sharing a temporary directory across your examples?

vext01:
Yeah. I think that should be possible. It would be awkward for the user to have to somehow figure out which temp directory to run a given test in when you've passed `test_dir()` many times.

Do you need to just create subdirs under the temp dir, so as to namespace them?

ltratt:
I bet the problem goes away with `--test-threads=1` right?

vext01:
I don't think i can apply that argument to an example, can I? `RUST_TEST_THREADS=1` makes no difference.

ltratt:
`cargo test -- --test-threads=1` or however you run the tests.

vext01:
I'm running the tests like:
```
cargo run --example rust_lang_tester
```

ltratt:
`cargo run --example rust_lang_tester -- --test-threads=1`

vext01:
that works without failure (albeit slowly).

It looks like races on identical filenames to me.

ltratt:
Yes, it's 99.9999% certain a race of filenames *but* it's a user mistake. You're probably copying `test_dir/x.y` to `/tmp/x.y` and asking it to run a command on that -- but since lang_tester (like cargo test) runs things in parallel, you're sometimes seeing it run two commands on the same file name. In other words, the test commands are flakey, and the user needs to fix them. [This race condition exists in lang_tester before this commit FWIW, though I admit that this commit makes it even easier to stumble into it.]

vext01:
I think you're right. The code just uses the file stem, which isn't unique.

ltratt:
OK, do you want to try another test or are you happy with this or do you think it takes a loaded gun and makes it more powerful or ... ?

vext01:
There are two solutions to the problem that I encountered:

 * use one temp dir for each test dir.
 * use one test dir, but sub-directories under it.

The `.test_cmds()` interface gives you the full path to a test file, and whlist I'm free to do as I wish with it, doing either of the above fixes will be awkward.

Both solutions would require the user to extract the test dir from the absolute path to the test file. For the first solution, the test dir would be the key to a `testdir -> tempdir` mapping. For the second solution, the sub-dir would be named the same as the test dir.

In my example getting back to the test dir is easy, as you only ever need to take the penultimate path element to get the test dir, but in general it may not be possible for variable-length test dirs.

One thing we could do is have the closure pass an additional argument: the test dir in which the source file was found. Then both solutions are (I think) easy, but it complicates the interface for users who only have one test dir.

Any thoughts?

ltratt:
The user is passed the full path so they can extract the directory if they want.

vext01:
I understand, but it's not ergonomic.

Suppose we have:
```
    .test_dir("a")
    .test_dir("b/a")
```

Suppose all three test dirs contain a test file called "sausage.c" and that our `run.rs` is in a sub-directory called `b`. So the directory tree is:

Now the closure for `test_cmds()` gives us the path:
```
/home/rms/my_proj/b/a/sausage.c
```

In this case in order to know that the first test dir is the parent of the file, you would have to inspect CARGO_MANIFEST_PATH.

It's not impossible, but it is annoying.

ltratt:
lang_tester is deliberately simple: we can't cope with *every* possibility without making the API horribly complex. If the user wants to use, for example, a temporary directory, IMHO it's their job to structure things correctly within it. We can never solve all possible problems for them.

So, let's move on, and just ask: is this PR a good idea or not?

vext01:
I'm not really satisfied with the new interface tbh.

I wonder if it'd just be easier to keep the old interface and if the user needs sub-dirs, then they can write a function to generate and run different lang tester instances with different test dirs.

vext01:
Yes, but as you said, it's user error and I don't see a good solution.

vext01:
If you still want to keep this change, I'd recommend you write a test which has a non-trivial set of  test_dirs all sharing the same temp dir. I think you will find the API awkward and limiting for the reasons outlined above.

PR #84, opened 2021-04-30
Prepare a 0.6.1 release

ltratt:
0.6.0 had a nasty bug whereby file filtering didn't work. I've yanked that from crates.io but we should prepare a correct version ASAP.


vext01:
Is it possible to add a test to prevent future regressions?

PR #85, opened 2021-05-02
Correct docs to change "failure" to "error"

icefoxen:
This matches what is in parser.rs


ltratt:
Good spot! Might you also be able to correct the same text (and mistake!) in `src/lib.rs` in this PR?

icefoxen:
Done, good catch.

ltratt:
Thank you!

bors r+

PR #87, opened 2021-05-23
Fix stdin close race.

ltratt:
If a test has a "stdin" clause, we close() stdin while still executing the program to stop it waiting forever for more input that will never come. However, that means that we closed both the "raw" file handle and dropped the `stdin`
variable which then (sort of silently) closed the file handle a second time. Since file handles are just integers and can be reused by the operating system, this led to a highly entertaining race where we closed (say) fd 6, subsequently fd 6 was reused (perhaps for another child processes's stdout), and then we dropped/closed fd 6 again -- even though it is in use elsewhere.

The fix for this is to `forget` the `stdin` variable if its been explicitly dropped.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -841,9 +841,9 @@ fn run_cmd(
         .spawn()
         .unwrap_or_else(|_| fatal(&format!("Couldn't run command {:?}.", cmd)));
 
-    let stdin = child.stdin.as_mut().unwrap();
-    let stderr = child.stderr.as_mut().unwrap();
-    let stdout = child.stdout.as_mut().unwrap();
+    let mut stdin = child.stdin.take().unwrap();
+    let mut stderr = child.stderr.take().unwrap();
+    let mut stdout = child.stdout.take().unwrap();
 
     let stdin_fd = stdin.as_raw_fd();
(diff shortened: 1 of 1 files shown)


src/tester.rs
```diff
@@ -907,9 +907,13 @@ fn run_cmd(
...
-                    // This is a bit icky, because the `stdin` variable will later close stdin when the
-                    // variable is dropped. However, some programs expect stdin to be closed before
-                    // they'll continue, so this is the least worst option.
+                    // If we don't close the child's stdin now, the child process might hang,
+                    // waiting forever for input that will never come. However, if we're not
+                    // careful, we'll close the underlying file handle twice: once here and once
+                    // when Rust drops the `stdin` variable. Since file handles can be reused that
+                    // means we could close a file handle that's now being used elsewhere. Thus
+                    // because we close the child's stdin here, we need to make sure we `forget`
+                    // the `stdin` variable at every return point from this function.
```
vext01, line 916: If you explicitly drop stdin here (maybe not possible?) instead of closing it, would we then need the forgets?

ltratt: It can't be dropped here, as Rust won't be able to prove that it's not used on the next iteration of the loop.

PR #91, opened 2021-06-16
Fix docs for, and rename, 'extra_arg'

ltratt:
The `extra-args` option is useful, but the documentation was misleading (https://github.com/softdevteam/lang_tester/commit/d014d43fb0f184746a37abb75af41b398e197fef) and indeed the very name of the key is misleading (https://github.com/softdevteam/lang_tester/commit/099096cb9936957be6d306d76593f9bf639775fa). The latter is a backwards incompatible change, but since I had to look at the source code to work out what's going on, it's hard to believe that many other people would be able to use this option correctly given it's currently-misleading name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,9 +143,10 @@ Each test command must define at least one sub-test:
 Test commands can alter the general command by specifying zero or more of the
 following:
 
-  * `extra-args: <arg 1> [... <arg n>]`, where each space separated argument
-    will be appended, in order, to those arguments specified as part of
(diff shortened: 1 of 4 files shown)


ltratt:
https://github.com/softdevteam/lang_tester/pull/91/commits/b2711714e6935c369e13dddcabb69951f9a50a9f will need squashing away.

ptersilie:
Please squash.

ltratt:
Squashed.

PR #100, opened 2022-01-14
enum-iterator-derive changed its license.

ltratt:


vext01:
Is there a way to turn the warnings into errors?

ltratt:
They are errors: what's happened, I think, is that, since we last ran `cargo deny` in buildbot.sh, the license in a dependency has changed. That's why it wasn't caught before: it didn't happen before!

vext01:
I can't seem to repro it any more, but the warning I had seen was:

```
warning[L006]: license was not encountered                                                                                                                                                         
  ┌─ /ci/deny.toml:6:5                                                                                                                                                                             
  │                                                                                                                                                                                                
6 │     "BSD-3-Clause",                         
  │     ^^^^^^^^^^^^^^ unmatched license allowance

licenses ok    
```

I think it's likely that this warning has been happening for some time, but since `cargo-audit` (presumably) returns non-zero, CI hasn't been catching it.

Why don't we add `-D warnings` to `cargo audit` to make warnings fail the CI?

vext01:
i meant to say `cargo-deny` in the above, not `cargo-audit`.

And `cargo-deny` doesn't have the ability to make warnings errors, it seems :\

Oh well.

ltratt:
So we can merge this then?

ltratt:
OK, so I had an old `Cargo.lock` which means this PR is unneeded. But I think the warning that triggered it is also based an old `Cargo.lock`. @vext01 I'm going to close this; if you get another warning about licenses, you might want to flush a `Cargo.lock` file on that repo.

PR #104, opened 2022-08-21
Add support for the `--list` flag.

### LINKED ISSUE #103: Support `--list` function
I use [nextest](https://nexte.st/) to test the programs in my project. I would like to integrate this with `lang_tester` (which I also use), however, this is currently not possible because `lang_tester` doesn't support [all of the features `nextest` requires](https://nexte.st/book/custom-test-harnesses.html#manually-implementing-a-test-harness).

It would be wonderful if these could be implemented (I would be happy to do this).

teymour-aldridge:
Fixes #103.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -355,10 +355,20 @@ impl LangTester {
                     "Number of threads used for running tests in parallel",
                     "n_threads",
                 )
+                .optflag("", "exact", "Print the test output exactly.")
+                .optflag("", "list", "List all the tests which can be run.")
+                .optopt("", "format", "Specify the output format", "This is currently silently ignored.")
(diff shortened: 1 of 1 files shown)


ltratt:
This seems a bit confusing to me: `--list` doesn't really do anything? I think for this to be considered we would at least have to document it with something like "Ignored for compatibility" to avoid raising expectations that the code can't actually satisfy :)

ltratt:
`cargo test X` with lang_tester will run tests matching `X` if that's what you mean? [Note that also includes tests AXB so long as they have `X` in them!]

ltratt:
I think this has timed out. Feel free to reopen it if there's new activity!

PR #106, opened 2023-01-23
Start rewriting the poll() loop.

ltratt:
This is shamelessly adapted from extsmaild.

@vext01 Could you try this in-progress PR and let me know what happens in your situation? You might just run into a `todo!()` (a couple of these imply changes elsewhere in lang_tester, but I haven't thought that through yet), but even that will might tell me something.

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -24,7 +24,8 @@ use std::{
 use fm::{FMBuilder, FMatchError};
 use getopts::Options;
 use libc::{
-    close, fcntl, poll, pollfd, F_GETFL, F_SETFL, O_NONBLOCK, POLLERR, POLLHUP, POLLIN, POLLOUT,
+    close, fcntl, poll, pollfd, F_GETFL, F_SETFL, O_NONBLOCK, POLLERR, POLLHUP, POLLIN, POLLNVAL,
+    POLLOUT,
 };
 use termcolor::{Color, ColorChoice, ColorSpec, StandardStream, WriteColor};
 use threadpool::ThreadPool;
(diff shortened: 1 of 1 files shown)


vext01:
Wow. Big change! I'll see if i can get a crash.

vext01:
Using this branch I have what appears to be a hangs. Do you want me to investigate?

vext01:
FWIW, the lang tester suite itself hangs too. That will be easier to repro/debug than the jit.

ltratt:
It doesn't hang for me, at least on OpenBSD... Unix, wot fun...

vext01:
I tested on Linux, so could be localised to that?

ltratt:
@vext01 The new commit can run `cargo test && cargo run --example fm_options && cargo run --example rust_lang_tester` on OpenBSD and Linux without hanging. Please see what it does to your yk stuff (there are still `todo`s and so on left, so don't expect miracles)!

vext01:
I'm afraid it still hangs.

ltratt:
@vext01 Please try the new commit. It has survived 1000 runs on Linux and OpenBSD on the test suite, so ... maybe it'll work!

ltratt:
@vext01 It sounds like this is maybe ready for squashing?

vext01:
I think so. I got 1150 iterations of the jit test suite out before hitting an unrelated issue, so I think lang_tester is fixed.

Please squash.

ltratt:
Squashed.

vext01:
Will you be making a new release after this?

PR #108, opened 2023-04-04
Add rerun-if-{status, stderr, stdout}

ltratt:
This PR adds support for rerunning tests if a) the test fails b) at least one of `rerun-if-{status, stderr, stdout}` matches.

Consider this test (which is included in the PR):

```python
# VM:
#   status: success
#   rerun-if-status: 42

import os, sys

cookie = os.path.join(os.environ["CARGO_MANIFEST_DIR"], "target", "tmp", "rerun_status_cookie")
i = 0
if os.path.exists(cookie):
    i = int(open(cookie, "r").read().strip()) + 1
    if i == 5:
        sys.exit(0)

open(cookie, "w").write(str(i))
sys.exit(42)
```

The first 4 times this test runs it will return with exit code 42, causing the test to be rerun. On the 5th time it will return with exit code 0.

The default number of times to rerun tests is arbitrarily set at 3, but users can override this.

Note that I have vaccillated several times over how to define this feature. At one point I thought it should be at the Rust-configuration level but, on reflection, that felt too fragile. If the user has lots of tests that should be rerun (e.g. if there's a network error) they're better off marking each one, since at least then forgetting to mark them as such will be spotted. In contrast, if there's a blanket application of rerun logic, it's more likely to accidentally cause errors to be suppressed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,11 @@ path = "examples/rust_lang_tester/run_tests.rs"
 name = "fm_options"
 path = "examples/fm_options/run_tests.rs"
 
+[[test]]
+name = "lang_tests"
+path = "lang_tests/rerun/main.rs"
+harness = false
+
 [dependencies]
 fm = "0.2"
 getopts = "0.2"
--- a/lang_tests/rerun/lang_tests/rerun_status.py
+++ b/lang_tests/rerun/lang_tests/rerun_status.py
@@ -7,10 +7,9 @@ import os, sys
 cookie = os.path.join(os.environ["CARGO_TARGET_TMPDIR"], "rerun_status_cookie")
 i = 0
 if os.path.exists(cookie):
-    i = int(open(cookie, "r").read().strip())
+    i = int(open(cookie, "r").read().strip()) + 1
     if i == 5:
         sys.exit(0)
-    i += 1
 
 open(cookie, "w").write(str(i))
 sys.exit(42)
--- a/lang_tests/rerun/lang_tests/rerun_status.py
+++ b/lang_tests/rerun/lang_tests/rerun_status.py
@@ -1,0 +1,16 @@
+# VM:
+#   status: success
+#   rerun-if-status: 42
+
+import os, sys
+
+cookie = os.path.join(os.environ["CARGO_TARGET_TMPDIR"], "rerun_status_cookie")
+i = 0
+if os.path.exists(cookie):
+    i = int(open(cookie, "r").read().strip())
+    if i == 5:
+        sys.exit(0)
+    i += 1
+
+open(cookie, "w").write(str(i))
+sys.exit(42)
--- a/lang_tests/rerun/lang_tests/rerun_stderr.py
+++ b/lang_tests/rerun/lang_tests/rerun_stderr.py
@@ -8,11 +8,10 @@ import os, sys
 cookie = os.path.join(os.environ["CARGO_TARGET_TMPDIR"], "rerun_stderr_cookie")
 i = 0
 if os.path.exists(cookie):
-    i = int(open(cookie, "r").read().strip())
+    i = int(open(cookie, "r").read().strip()) + 1
     if i == 5:
         sys.stderr.write("a")
         sys.exit(0)
-    i += 1
 
 open(cookie, "w").write(str(i))
 sys.stderr.write("b")
--- a/lang_tests/rerun/lang_tests/rerun_stderr.py
+++ b/lang_tests/rerun/lang_tests/rerun_stderr.py
@@ -1,0 +1,18 @@
+# VM:
+#   status: success
+#   stderr: a
+#   rerun-if-stderr: b
+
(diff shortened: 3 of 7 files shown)


examples/rerun/lang_tests/rerun_stdout.py
```diff
@@ -0,0 +1,18 @@
...
+#   status: success
+#   stdout: a
+#   rerun-if-stdout: b
+
+import os, sys
+
+cookie = os.path.join(os.environ["CARGO_MANIFEST_DIR"], "target", "tmp", "rerun_stdout_cookie")
+i = 0
+if os.path.exists(cookie):
+    i = int(open(cookie, "r").read().strip())
```
vext01: Are we OK with having two file handles to the cookie open concurently? Should we use a context manager to close the first? No idea.

examples/rerun/run_tests.rs
```diff
@@ -0,0 +1,61 @@
...
+use std::{
+    env,
+    fs::{read_to_string, remove_file},
+    path::PathBuf,
+    process::Command,
+};
+
+/// In order that we can meaningfully have the same test behave differently from run to run, we
+/// have to leave cookies behind in target/tmp, which we must clear before starting a test run.
+/// This is horrible, but without doing something stateful, we can't test anything.
```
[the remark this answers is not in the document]
ltratt: PID reuse probably kills that idea, and you'd still need a cookie file, surely?

examples/rerun/run_tests.rs
```diff
@@ -0,0 +1,61 @@
...
+    "rerun_stdout_cookie",
+];
+
+use lang_tester::LangTester;
+use regex::Regex;
+
+fn main() {
+    for cookie in COOKIES {
+        let mut p = PathBuf::new();
+        p.push(env::var("CARGO_MANIFEST_DIR").unwrap());
```
vext01, line 25: It doesn't hugely matter, but do you know the `[elems, ...].collect::<PathBuf>()` trick?

examples/rerun/lang_tests/rerun_stdout.py
ltratt: > Are we OK with having two file handles to the cookie open concurently? 

We aren't! That's why each test uses a separate cookie file: without that, the three tests would fail in weird ways.

examples/rerun/run_tests.rs
ltratt: I didn't!

examples/rerun/run_tests.rs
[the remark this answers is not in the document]
vext01: > PID reuse probably kills that idea

If the PID is immediately re-used, we have an infinite loop. Linux uses monotonic PIDs, OpenBSD random...

> and you'd still need a cookie file, surely?

I meant something like, don't rerun if `pid % 3 == 0`.

But I don't like this. Your way is better.

examples/rerun/run_tests.rs
vext01: It's not always practical, but often makes building paths a one-liner.

ltratt: Just tried this, and it turns 5 lines into 8 lines (since `iter` is also needed) so not a win this case it seems. Oh well, it's a useful trick to know!

vext01: Oh well.

examples/rerun/lang_tests/rerun_stdout.py
ltratt: Cookie files aren't opened concurrently.

vext01:
```
open(cookie, "r")
```

with no corresponding `.close()`, and then:

```
open(cookie, "w")
```

later.

What am I missing?

ltratt: I see what you mean! Due to the wonders of reference counting, those won't be concurrent. And, even if they were, the write would (correctly) create a new inode which wouldn't interfere with the read. So we're OK either way.

vext01: I did not know that Python closed files when the ref count drops to zero! Begs the question: why to use `with open() as f: ...`

>  even if they were, the write would (correctly) create a new inode which wouldn't interfere with the read

I don't think that's right, but it's also irrelevant now.

vext01:
Ready to merge then?

ltratt:
Yes.

ltratt:
bors try

ltratt:
bors try

ltratt:
bors try

ltratt:
OK, this finally interacts properly with `cargo test` which a) defines `CARGO_TARGET_TMPDIR` (which is nice!) b) only defines it at compile-time.

This needs some heavy squashing if you're OK with it!

vext01:
LGTM.

Please squash!

ltratt:
Squashed.

Issue #110, opened 2023-09-21
namespaced test filtering doesn't work with lang_tester.


vext01:
I've just noticed that namespaced filtering doesn't work with lang_tester.

When you run `cargo test` it prints the full namespaced path of the tests like:
```
...
test compile::jitc_yk::aot_ir::tests::string_deser ... ok                                                             
test compile::jitc_yk::aot_ir::tests::deser_and_display ... ok                                                        
test mt::tests::dont_trace_two_locations_simultaneously_in_one_thread ... ok                                          
test mt::tests::basic_transitions ... ok
...
```

And you can do (e.g.):
```
$ cargo test mt::tests::basic_transitions
...
running 1 test
test mt::tests::basic_transitions ... ok
...
```

But this doesn't work for tests driven by lang_tester. E.g. if running `cargo test` with no args prints:
```
$ cargo test
...
Running C tests with opt level -O0...

running 75 tests
test lang_tests::many_threads_one_loc.c ... ignored
test lang_tests::missing_control_point.c ... ok
test lang_tests::intrinsics.c ... ok
test lang_tests::mutable_global.c ... ok
...
```

We should be able to pluck one of those tests in isolation with:
```
$ cargo test lang_tests::missing_control_point.c
...
Running C tests with opt level -O0...

running 0 tests
...
```

This is a bit annoying for Yk development, where I want to test only `lang_tests::simple.c`, but filtering by `simple.c` runs multiple tests with similar names.

PR #111, opened 2023-09-21
Use the full test name (e.g. `a::b::c`) when filtering.

ltratt:
Previously we just filtered on the leaf, so if you had multiple tests with the same leaf but in different directories, you could not filter them individually.

Fixes https://github.com/softdevteam/lang_tester/issues/110 (or hopefully does! @vext01 does this work for you?).

### DIFF
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -331,12 +331,18 @@ impl LangTester {
             // If the user has named one or more tests on the command-line, run only those,
             // filtering out the rest (counting them as ignored).
             .filter(|x| {
-                let x_path = x.to_str().unwrap();
+                let test_fname = format!(
+                    "lang_tests::{}",
+                    test_fname(
+                        &self.inner.test_dir.as_ref().map(|x| x.as_path()).unwrap(),
+                        &x.as_path(),
+                    )
+                );
                 match self.cmdline_filters.as_ref() {
                     Some(fs) => {
                         debug_assert!(self.use_cmdline_args);
                         for f in fs {
-                            if x_path.contains(f) {
+                            if test_fname.contains(f) {
                                 return true;
                             }
                         }


vext01:
I don't know if this is related, but the suite name (before the `::`) isn't necessarily a directory name. In `yk` the tests live in the `tests` crate, but the suite is `lang_tests`.

ltratt:
At the moment the `lang_test::` prefix is a sort-of hack, but you're right that exposing that to filtering would make sense. See if e265bc3 helps.

vext01:
That works.

As a separate feature request (which may not be worth your time), in yk, the tests directory has several lang_tester suites, which all get smooshed into the `lang_tests` namespace. Ideally the suites would have a unique sub-prefix (the langtester `test_dir`?).

Another upshot of the smooshing, if two suites have tests with the same name, they cannot be uniquely identified.

E.g.

```
tests/suite1/blah.c
tests/suite2/blah.c
```

Are both assigned `lang_tests::blah.c`, but it would be better if they were: `lang_tests::suite1::blah.c` and `lang_tests::suitie2::blah.c`.

I willing to bet that what I'm asking for is a huge hassle, so don't worry if it's a lot of work.

ltratt:
>  Ideally the suites would have a unique sub-prefix (the langtester test_dir?).

This might be harder. If we definitely need it, I can look into it, but if it's just a theoretical possibility, I might punt on it!

OK to squash?

vext01:
> If we definitely need it

**need** would be a bit strong here. In `yk` no test can be named the same at the moment, as all the C tests have a unique `.c` prefix.

Maybe we should raise an issue and wait for a complaint? :laughing:

ltratt:
> Maybe we should raise an issue and wait for a complaint? 😆

You can... I'll forget :p If it becomes a problem, I suggest shouting at that point. I tend to think of lang_tester as in the no-mans-land between "maintained" and "unmaintained"!

OK to squash?

ltratt:
Squashed.

PR #118, opened 2024-01-22
Add `test_path_filter` and deprecate `test_file_filter`.

ltratt:
`test_file_filter` filters out non-file things, which makes it impossible to, for example, use directories as tests.

Rather than change the semantics of `test_file_filter`, this commit adds `test_path_filter` which doesn't filter out non-file things, but is otherwise identical to `test_file_filter`. It also marks `test_file_filter` as deprecated, with a generic way to migrate code to `test_path_filter`, allowing users to mechanically migrate their code.


ltratt:
@jacob-hughes I always forget that "examples" aren't run by `cargo test`! Fixed in 4f8de4d.

jacob-hughes:
LGTM please squash

ltratt:
Squashed.

PR #120, opened 2024-01-30
Replace `ignore` with `ignore-if`.

ltratt:
This PR's main aim is to remove `ignore` and add `ignore-if: <cmd>` which allows a test to flexibly specify when it should/shouldn't be ignored (https://github.com/softdevteam/lang_tester/commit/9ffa2b2f02c8f51de722bcc5b64c8c9ad642729a). However, this means that the ability to specify a human string for *why* a test is ignored is lost so this PR also adds the ability to allow comment lines (https://github.com/softdevteam/lang_tester/commit/d09c0dc05a59dd7fb9cb58ab68fe8f2485ad5bd9).

My guess is that the main way this will be used is with `test` so you might have things like `ignore-if: test $YKB_TRACER == "swt"` or `ignore-if: test $YKB_TRACER != "swt"` and so on. @Pavel-Durov does this allow you to specify tests in your yk PR in the way you need?

This is a breaking change, so if we do think this is useful and fits in with lang_tester's design, then it does imply a breaking release.

### DIFF
--- a/examples/rust_lang_tester/lang_tests/ignore.rs
+++ b/examples/rust_lang_tester/lang_tests/ignore.rs
@@ -1,4 +1,4 @@
-// ignore: this test is intentionally ignored
+// ignore-if: true
 // Compiler:
 //   status: success
 
--- a/examples/rust_lang_tester/lang_tests/ignore.rs
+++ b/examples/rust_lang_tester/lang_tests/ignore.rs
@@ -1,4 +1,5 @@
-// ignore-if: true
+// # Always ignore this test
+// ignore-if: echo 123 | grep 2
 // Compiler:
 //   status: success
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -159,11 +159,10 @@
 //! its `stderr` output should warn about an unused variable on line 12; and the resulting binary
 //! should succeed produce `Hello world` on `stdout`.
 //!
-//! A file's tests can be ignored entirely if a test command `ignore` is defined:
+//! A file's tests can be ignored entirely with:
 //!
-//!   * `ignore: [<string>]` specifies that this file should be ignored for the reason set out in
-//!     `<string>` (if any). Note that `<string>` is purely for user information and has no effect
-//!     on the running of tests.
+//!   * `ignore-if: <cmd>` defines a shell command that will be run to determine whether to ignore
+//!     this test or not. If `<cmd>` returns 0 the test will be ignored, otherwise it will be run.
 //!
 //! `lang_tester`'s output is deliberately similar to Rust's normal testing output. Running the
 //! example `rust_lang_tester` in this crate produces the following output:
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -138,21 +138,21 @@
 //!   * `exec-arg: <string>` specifies a string which will be passed as an additional command-line
 //!     argument to the command (in addition to those specified by the `test_cmds` function).
 //!     Multiple `exec-arg`s can be specified, each adding an additional command-line argument.
-//!   * `stdin: <string>`, text to be passed to the command's `stdin`. If the command exits without
-//!     having consumed all of `<string>`, an error will be raised. Note, though, that operating
-//!     system file buffers can mean that the command *appears* to have consumed all of `<string>`
-//!     without it actually having done so.
+//!   * `stdin: <string>` specifies text to be passed to the command's `stdin`. If the command
+//!     exits without consuming all of `<string>`, an error will be raised. Note, though, that
+//!     operating system file buffers can mean that the command *appears* to have consumed all of
+//!     `<string>` without it actually having done so.
 //!
 //! Test commands can specify that a test should be rerun if one of the following (optional) is
 //! specified and it matches the test's output:
 //!
-//!   * `rerun-if-status`: follows the same format as the `status`.
-//!   * `rerun-if-stderr` and `rerun-if-stdout`: follow the same format as `stderr` and `stdout`.
+//!   * `rerun-if-status` follows the same format as the `status`.
+//!   * `rerun-if-stderr` and `rerun-if-stdout` follow the same format as `stderr` and `stdout`.
 //!
 //! These can be useful if tests are subject to intermittent errors (e.g. network failure) that
 //! should not be considered as a failure of the test itself. Test commands are rerun at most *n*
 //! times, which by default is specified as 3. If no `rerun-if-` is specified, then the first time
-//! a test fails, it will be reported to the users.
+//! a test fails, it will be reported to the user.
 //!
 //! The above file thus contains 4 meaningful tests, two specified by the user and two implied by
 //! defaults: the `Compiler` should succeed (e.g. return a `0` exit code when run on Unix), and
@@ -161,7 +161,7 @@
 //!
 //! A file's tests can be ignored entirely if a test command `ignore` is defined:
 //!
-//!   * `ignore: [<string>]`, specifies that this file should be ignored for the reason set out in
+//!   * `ignore: [<string>]` specifies that this file should be ignored for the reason set out in
 //!     `<string>` (if any). Note that `<string>` is purely for user information and has no effect
 //!     on the running of tests.
 //!
--- a/src/tester.rs
+++ b/src/tester.rs
@@ -61,6 +61,7 @@ struct LangTesterPooler {
     test_threads: usize,
     ignored: bool,
     nocapture: bool,
+    comment_prefix: Option<String>,
     test_extract: Option<Box<dyn Fn(&Path) -> String + RefUnwindSafe + Send + Sync>>,
     fm_options: Option<
         Box<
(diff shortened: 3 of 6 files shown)


Pavel-Durov:
@ltratt that's perfect for my use case in YK PR!

examples/rust_lang_tester/lang_tests/ignore.rs
```diff
@@ -1,4 +1,4 @@
-// ignore: this test is intentionally ignored
```
vext01, line 1: So do we lose the ability to say why a test is ignored then?

I think forcing a reason was a good thing. I can barely remember why I did things last week, let alone why someone else did something 2 years ago :)

src/lib.rs
```diff
@@ -159,11 +159,11 @@
 //! its `stderr` output should warn about an unused variable on line 12; and the resulting binary
 //! should succeed produce `Hello world` on `stdout`.
 //!
-//! A file's tests can be ignored entirely if a test command `ignore` is defined:
+//! A file's tests can be ignored entirely with:
 //!
-//!   * `ignore: [<string>]` specifies that this file should be ignored for the reason set out in
-//!     `<string>` (if any). Note that `<string>` is purely for user information and has no effect
-//!     on the running of tests.
+//!   * `ignore-if-cmd: <cmd>` defines a shell command that will be run to determine whether to
```
vext01, line 164: Should `ignore-if-cmd` be `ignore-if` or are they two separate facilities?

ltratt: Oops, that's a typo from an earlier design. Fixed in 679ac45.

examples/rust_lang_tester/lang_tests/ignore.rs
ltratt: See https://github.com/softdevteam/lang_tester/pull/120/commits/d09c0dc05a59dd7fb9cb58ab68fe8f2485ad5bd9 -- this PR makes things much more flexible!

ltratt: That's a human/per-project problem, not a lang_tester problem :)

vext01:
Based on my preliminary look, we've added something very general: a shell command.

My worry is that forking will slow down the test suite.

For now, we only need the ability to ignore something based on environment variables.

I wonder if `ignore-if` should be `ignore-if-env`, e.g.:
```
// ignore-if-env: USER=edd
```

If you needed to skip tests based on the output of shell commands (will we need it? I don't know), it would be doable, but more effort. You'd have to get the test runner to run the shell command and set an env var.

That then raises the question: should there be `ignore-if-env` and `ignore-if-cmd`?

(I bet that was your earlier design?)

ltratt:
I don't think `ignore-if-env` is general enough because you can't specify things like "ignore this test with opt-level > 1`. I almost added an expression parser in but... given that lang_tester is already running *your* program as a sub-command, firing a shell with `test` is going to be a manageable cost IMHO.

vext01:
> I don't think ignore-if-env is general enough because you can't specify things like "ignore this test with opt-level > 1".

If knowing the opt level requires a shell-out, tou could do it by getting your runner (e.g. `langtest_c.rs`) to shell out and set an `OPT_LEVEL` env var based on the result. It's more work for the user, but the one shell-out can be shared for all tests.

> given that lang_tester is already running your program as a sub-command, firing a shell with test is going to be a manageable cost IMHO.

I could buy that, so if you still agree that the current design is acceptable, I'll do a deeper review. Let me know.

ltratt:
I started with `ignore-if-env` and ended up with this PR so I'm more a fan of this style, so if you buy the argument, please go ahead with a review!

examples/rust_lang_tester/lang_tests/ignore.rs
vext01: I'd argue that your change made it a human/per-project problem. Before you had to provide a reason. No ifs, not buts.

Perhaps if you use `ignore-if`, you have to also provide a `ignore-reason` line too?

vext01:
I think we should add a test that uses a `ignore-if` that is more than a simple `true`/`false` to check that arguments are passed to the shell properly.

examples/rust_lang_tester/lang_tests/ignore.rs
ltratt: `ignore: <reason>` always had `<reason>` as optional, so this PR doesn't change the optionality. I don't want to mandate how people use this: indeed, the comment syntax is far more flexible than what we had before IMHO.

ltratt:
> I think we should add a test that uses a ignore-if that is more than a simple true/false to check that arguments are passed to the shell properly.

Any thoughts on what?

src/tester.rs
```diff
@@ -128,6 +130,25 @@ impl LangTester {
         self
     }
 
+    /// If set, defines what top-level lines will be treated as comments if they begin with
```
vext01, line 133: top-level is confusing me a bit here.

If you look at a typical lang_test, it looks like:
```
// Compiler:
//   stdout:
//     match me!
```

And now we are adding the ability to have an inner comment, so a test can look like:
```
// Compiler:
//   stdout:
//     # flibble
//     match me!
```

So when you say "defines what top-level lines will be treated as comments", I'm thinking of the top-level `//` comments, not the `#' ones.

Do you see my concern?

examples/rust_lang_tester/lang_tests/ignore.rs
vext01: I didn't realise that it was optional, so I'll back down.

vext01:
> Any thoughts on what?

Huh. I didn't ask for any thoughts :P

src/tester.rs
ltratt: If you can think of a better phrasing, I'm all ears, but I think this comes under the heading of "this is how lang_tester works and you can't avoid knowing a bit about how a) your language parses comments b) your runner extracts test c) how lang_tester parses things".

vext01: I've just realised something else. By top-level, you mean it must appear at the same level as (e.g.) `Compiler:`.

So in fact my example above is invalid. But should it be? I've often wanted to comment why I'm matching something, so perhaps comments should be allowed anywhere, and then all of this "top-level" talk disappears?

vext01: It's true that you might accidentally comment something you want to match.

I'll leave the decision up to you. I'm aware that I've been pedantic/annoying enough for one PR!

ltratt: On reflection I agree that generalising this is probably a good idea. Try 2223188.

src/tester.rs
```diff
@@ -130,17 +130,8 @@ impl LangTester {
...
-    /// For example, if we set `comment_prefix` to `#` then:
-    ///
-    /// ```text
-    /// # Considered a top-level comment by lang_tester
-    /// Compiler:
-    ///   stdout:
-    ///     # A literal string beginning with `#`
-    /// ```
+    /// If set, defines what lines will be treated as comments if, ignoring the current level of
+    /// indentation, they begin with `comment_prefix`.
```
vext01, line 134: This all LGTM. I wonder if here we should say:

> If set, defines what lines of a lang_tester test will be treated as comments if, ignoring the current level of indentation, they begin with `comment_prefix`.
>
> Note that this is not to be confused with another language's comment lines being in a test file when extracting the lang_tester test (e.g. using [LangTester::take_while].

Second sentence is iffy, but I hope you can see what I'm aiming at?

vext01: Indeed. It's difficult to put into words.

Maybe we should just move on and wait for a user to squeal?

ltratt: > Maybe we should just move on and wait for a user to squeal?

I tend to agree.

vext01:
I think it's just [this test](https://github.com/softdevteam/lang_tester/pull/120#issuecomment-1916625618) outstanding now?

ltratt:
> I think it's just https://github.com/softdevteam/lang_tester/pull/120#issuecomment-1916625618 outstanding now?

I'm still awaiting an idea of what one could actually do here from you! The problem is that one is not in control of the environment, so it's hard to know what (say) an environment variable that's always set (or not) might be. I was astonished recently to find situations where even `$SHELL` isn't set on some systems! For example, I wonder if `$PATH` is set on Windows or not?

vext01:
I don't mean to test environment variables. I mean to test some shell command with more than a single argument, e.g.:

```
ignore-if: echo "123" | grep 2
```

This should skip the test, as grep exits 0 on a match.

(I want to check that `echo "123" | grep 2` isn't interpreted as a single argument with spaces inside)

(obviously guard such a test for unix platforms)

ltratt:
>  (obviously guard such a test for unix platforms)

How?

ltratt:
@vext01 Actually, I can't be bothered guarding for Unix, so d45f9aa adds tests (and fixes an issue where `ignore-if` splats on the terminal) that are Unix only --- but I suspect even Windows might pass this test these days!

vext01:
Great! thanks.

And sorry (not sorry) for being persistent :)

Please squash.

ltratt:
Squashed.

PR #123, opened 2024-02-25
Fix example

ltratt:
The example code in both `lib.rs` and `README.md` is both slightly out-of-date and slightly confusing. This PR fixes a few obvious issues.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,7 +9,7 @@
 //! [`compiletest_rs`](https://crates.io/crates/compiletest_rs), looks as follows:
 //!
 //! ```rust,ignore
-//! use std::{fs::read_to_string, path::PathBuf, process::Command};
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -9,7 +9,7 @@
 //! [`compiletest_rs`](https://crates.io/crates/compiletest_rs), looks as follows:
 //!
 //! ```rust,ignore
-//! use std::{fs::read_to_string, path::PathBuf, process::Command};
+//! use std::{env, fs::read_to_string, path::PathBuf, process::Command};
```
vext01, line 12: Is there a reason it's marked `ignore`? We could have caught this earlier?

ltratt: I'm not sure I want my doc tests spinning up arbitrary commands. But maybe I'm over-thinking it!

vext01: I'll merge this as-is, and you can have a think about it.

